Issue #19, opened 2017-08-16
Finish configurable weight system


cam72cam:
* Add weight to json files (in the properties section)
* Calculate the weight dynamically in MovableRollingStock based on cargo

PR #22, opened 2017-09-02
Added weight to the locos, tenders, and cars

StuffedCroc:
updated JSON files with realistic weights and updated EntityRollingStock to pull the weight from the JSONs.

in response to #19


cam72cam:
Looks Good :+1:

PR #300, opened 2018-04-21
Improved orthography

clerie:
I improved orthography in the German lang file

Especially 'Kupplung':
https://de.wikipedia.org/wiki/Kupplung_(Bahn)


cam72cam:
Thanks this'll be in the next version!

PR #397, opened 2018-08-06
support tenders. The 1% object will now be rendered if there is any i…

FriedrichLP:
…tem in the wagon (even just one)


cam72cam:
There's probably a better way of doing this.  Try moving this functionality into a already existing common base class

PR #400, opened 2018-08-07
Merge Railroad Crossing

FriedrichLP:
This branch adds a Railroad Crossing as a new track type you can select in the track blueprint


cam72cam:
Can you add a description of what this change adds?

FriedrichLP:
I added a description. Short but should be enough

PR #402, opened 2018-08-09
Lockable stock

### LINKED ISSUE #324: User lock states for stock.
To have the ability to lock stock from:
   -Other players breacking and entering your stock
   -Other players breacking your stock
   -No lock

It may also be usefull to give access to specific players.
2 ideas to do this from me:
   -Right click on stock with a nametag renamed to a username (no nametag consumption)
     (Second time to remove player)
   -IR specific friends list via commands

A tool to view who owns and has access a piece of rollingstock outside of OC
Specificly usefull for administration wanting to know who left there consist half on a switch and logged out. >:(

FriedrichLP:
Fixes #324


src/main/java/cam72cam/immersiverailroading/entity/EntityRollingStock.java
```diff
@@ -69,6 +89,37 @@ public String getDefinitionID() {
 		return this.defID;
 	}
 	
+	public int getLockType () {
+		return this.getDataManager().get(LOCK_TYPE);
```
cam72cam: LockType should be an Enum

src/main/java/cam72cam/immersiverailroading/entity/EntityRollingStock.java
```diff
@@ -103,6 +158,7 @@ protected void writeEntityToNBT(NBTTagCompound nbttagcompound) {
 		nbttagcompound.setString("defID", defID);
 		nbttagcompound.setDouble("gauge", gauge.value());
 		nbttagcompound.setString("tag", tag);
+		nbttagcompound.setInteger("lock_type", getLockType());
```
cam72cam: You forgot to save the lock owner

src/main/java/cam72cam/immersiverailroading/entity/EntityRidableRollingStock.java
```diff
@@ -110,12 +110,13 @@ public boolean processInitialInteract(EntityPlayer player, EnumHand hand) {
 		} else if (player.isRiding() && player.getRidingEntity().getPersistentID() == this.getPersistentID()) {
 			return false;
 		} else {
-			if (!this.world.isRemote) {
+			if (!this.world.isRemote && (getLockType() != 2 || getLockOwner().equals(player.getUniqueID())) || playerHasOp(player)) {
```
cam72cam: I see this added check in a few places.  Add a function on the rolling stock:
hasPermission(player, isBreaking)

Remove all the copy pasted conditionals

cam72cam: Also this change does not work.  The client will get False even when the server get's true.

Take a look at the original code and see that both the client and server side always get the same answer

src/main/java/cam72cam/immersiverailroading/entity/EntityRollingStock.java
```diff
@@ -127,7 +184,14 @@ protected void entityInit() {
 	
 	@Override
 	public boolean processInitialInteract(EntityPlayer player, EnumHand hand) {
-		return false;
+		if (player.getHeldItem(hand).getItem() == IRItems.ITEM_LOCK_KEY && (getLockType() == 0 || getLockOwner().equals(player.getUniqueID())) || playerHasOp(player)) {
```
cam72cam: Again, use the hasPermission function

src/main/java/cam72cam/immersiverailroading/entity/EntityRollingStock.java
```diff
@@ -158,7 +222,7 @@ public boolean attackEntityFrom(DamageSource damagesource, float amount) {
 		
 		if (damagesource.getTrueSource() instanceof EntityPlayer && !damagesource.isProjectile()) {
 			EntityPlayer player = (EntityPlayer) damagesource.getTrueSource();
-			if (player.isSneaking()) {
+			if (player.isSneaking() && (getLockType() == 0 || getLockOwner().equals(player.getUniqueID())) || playerHasOp(player)) {
```
cam72cam: hasPermission fn

FriedrichLP:
I will change everything in here as I got a new request for a lock GUI. I forgot to close the PS sorry xD

src/main/java/cam72cam/immersiverailroading/entity/EntityRollingStock.java
```diff
@@ -42,7 +43,7 @@
 
 public abstract class EntityRollingStock extends Entity implements IEntityAdditionalSpawnData {
 	
-	private static DataParameter<Integer> LOCK_TYPE = EntityDataManager.createKey(EntityRollingStock.class, DataSerializers.VARINT);	//0=everyone can enter/break, 1=everyone can enter, 2=no one can enter/break
+	private static DataParameter<String> LOCK_TYPE = EntityDataManager.createKey(EntityRollingStock.class, DataSerializers.STRING);	//0=everyone can enter/break, 1=everyone can enter, 2=no one can enter/break
```
cam72cam: This should remain an integer synchronizer.  Use the ordinal value of the enum instead

src/main/java/cam72cam/immersiverailroading/entity/EntityRidableRollingStock.java
```diff
@@ -110,11 +111,10 @@ public boolean processInitialInteract(EntityPlayer player, EnumHand hand) {
 		} else if (player.isRiding() && player.getRidingEntity().getPersistentID() == this.getPersistentID()) {
 			return false;
 		} else {
-			if (!this.world.isRemote) {
+			if (!this.world.isRemote && (getLockType() != LockType.LOCKED || getLockOwner().equals(player.getUniqueID())) || playerHasOp(player)) {
```
cam72cam: Should use the hasPermission function

src/main/java/cam72cam/immersiverailroading/entity/EntityRollingStock.java
```diff
@@ -69,6 +92,30 @@ public String getDefinitionID() {
...
+		return UUID.fromString(this.getDataManager().get(LOCK_OWNER));
+	}
+	
+	public void setLockOwner(String playerId) {
+		this.getDataManager().set(LOCK_OWNER, playerId);
+	}
+	
+	public void switchLockType (EntityPlayer player) {
+		setLockType(getLockType().next());
+		setLockOwner(EntityPlayer.getUUID(player.getGameProfile()).toString());
```
cam72cam: player has a get persistent ID function.  Use that instead of this stuff

src/main/java/cam72cam/immersiverailroading/entity/EntityRollingStock.java
```diff
@@ -78,6 +125,14 @@ public void onUpdate() {
 			}
 		}
 	}
+	
+	public boolean playerHasOp (EntityPlayer player) {
+		return  FMLCommonHandler.instance().getMinecraftServerInstance().getPlayerList().getOppedPlayers().getEntry(player.getGameProfile()) != null;
```
cam72cam: Is this safe on both the client and server side?

FriedrichLP: it works in single player but I've never tested it on an actual server

src/main/java/cam72cam/immersiverailroading/entity/EntityRollingStock.java
```diff
@@ -69,6 +92,30 @@ public String getDefinitionID() {
...
+		}
+		return UUID.fromString(this.getDataManager().get(LOCK_OWNER));
+	}
+	
+	public void setLockOwner(String playerId) {
+		this.getDataManager().set(LOCK_OWNER, playerId);
+	}
+	
+	public void switchLockType (EntityPlayer player) {
+		setLockType(getLockType().next());
```
cam72cam: Send the player a message when this happens so they know what lock level this is

cam72cam:
@FriedrichLP Create a new class EntityLockableRolling stock and insert it between EntityRollingStock and EntityBuildableRollingStock.

This will allow you to have all of the locking logic in one place.  It also will allow you to prevent other users from dissassembling your loco it it's locked.

Endalion:
Coupling, and paint brush should be owner-locked, is a separate lock type for inventory access/livestock loading ideal or is SEMI-LOCKED enough?

FriedrichLP:
Will be reworked later

PR #404, opened 2018-08-11
reduce Vec2f and String allocations in model.obj

asiekierka:
The Vec2f allocation reduction saves ~50MB, the String interning saves ~200KB. The amount of non-zero UV offset Vec2f is so small as to not be worth reducing further.

Unfortunately, the mod's models still consume about 200MB even with the reduction. The format should probably be optimized for memory usage further.


cam72cam:
Thanks!

The next step is probably to delete the models from memory once the have been uploaded to the GPU

asiekierka:
Yes, that would be a good idea.

If that happens, you might want to remove the string interning. Keep the Vec2f allocation reduction, though.

Issue #302, opened 2018-04-26
Support Thermal Expansion Fuels


cam72cam:

cam72cam:
Thanks @EweLoHD

PR #417, opened 2018-08-20
Added Support for Thermal Foundation/Expansion Fuels

EweLoHD:
Added Support for "Naphtha" and "Refined Fuel". Are the values that I gave them good?

--> #302

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/Config.java
+++ b/src/main/java/cam72cam/immersiverailroading/Config.java
@@ -19,6 +19,9 @@ public class Config {
 		@Comment({ "Enable Boiler Explosions" })
 		public static boolean explosionsEnabled = true;
 
+		@Comment({ "Enable environmental damage of Boiler Explosions"})
+		public static boolean explosionEnvDamageEnabled = true;
+		
 		@Comment({ "km/h to damage 1 heart on collision" })
 		public static double entitySpeedDamage = 10;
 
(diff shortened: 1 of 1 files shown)


FriedrichLP:
that should be already possible since you can add custom diesel fuels in the config. Maybe you can add them to the default fuel map

src/main/java/cam72cam/immersiverailroading/Config.java
```diff
@@ -19,6 +19,9 @@
 		@Comment({ "Enable Boiler Explosions" })
 		public static boolean explosionsEnabled = true;
 
+		@Comment({ "Enable environmental damage of Boiler Explosions"})
+		public static boolean explosionEnvDamageEnabled = true;
+
```
cam72cam: Why is this in here?

src/main/java/cam72cam/immersiverailroading/Config.java
```diff
@@ -101,7 +104,8 @@
 			Items.EMERALD.getRegistryName().toString()
 		};
 		
-		@Comment("Fuels for diesel Locomotives")
+		@Comment("Fuels for diesel Locomotives" +
```
cam72cam: What does the \n look like in the in-game UI and in the config text?

EweLoHD: image

src/main/java/cam72cam/immersiverailroading/Config.java
EweLoHD: idk, this is not new. Had some Problems with Github Desktop recently

src/main/java/cam72cam/immersiverailroading/Config.java
EweLoHD: and it doesn't show me comments in the in-game UI

Issue #217, opened 2018-02-10
Have steam locomotives generate steam as fluid


WrongWeekend:
Sometime in the future, IR steam locomotives would have better compatibility with other mods that use steam as a fluid if they generate and use the fluid themselves. This would potentially allow for certain mechanics such as using external boilers to quickly start up steam locomotives, using steam locomotives as stationary power sources, fireless steam locomotives, and even the use of steam heating for older passenger cars from equipped steam, diesel, and electric locomotives. The latter could involve things such as HP regen effects or possible Tough as Nails compatibility.

SpudsMcCat:
so irl  large roundhouses would have a steam generator, and the locos would get plugged into it over night (the exapnsion/contraxtion of  cooling down to ambient puts a ton of stress on the locos)  be cool if could do that, so have engines  warmed up and ready to go, w/o having to refuel each one seperattly

CarterlikesIR:
sounds cool

Raeffi:
there are also small steam locomotives for shunting that get filled with high pressure steam once and then run for a few hours

PR #422, opened 2018-08-30
Added Steam- Generator, Fluid and Augment [W.I.P, needs Review]

EweLoHD:
I started working on #217 , this is my progress:

- [x] Added new Fluid (/Gas): **Steam**
- [x] Added a new Multiblock: **Steam Generator**
  -  If it has Energy (inputted on the top) and Water (inputted on the right or left side), it produces Steam (outputted on the front side)
  - Uses 32 RF/tick Energy
  - Uses 10mB Water and produces 10mB Steam per Tick
  - Thanks to @trainman264 for this amazing Model
- [x] Added new Augments: **Steam Loader/Unloader**
  - If the Steam Loader has Steam (in its Augment Tank) it increases the Pressure of the Locomotive above it (if it's a Steam Locomotive)

  - Uses 100mB Steam and increases the Pressure by 0,05
  - Steam Unloader doesn't work yet (See _To-Do_)

### To-Do (W.I.P)
- [x] Get Steam Unloader to work
- [x] Allow that Water can be inputted from both Sides
- [x] Support Steam of other Mods
- [x] Add NBT Tags
- [x] Fix Steam Texture
- [x] Fix Steam Locomotive Condensing

### Need Review 
 - beside everything from my PR:
 - Steam Fluid (The Texture doesn't load. IDK why. I tried everything, but Steam still has these purple/black Texture)
 - Is there something that I forgot on my To-Do List?

Why am I already making a PR if I'm not completely done yet? Well, it's already playable, so it could be merged. My code needs review, because of the fluid texture and because my code is usually not the best. And mainly because I will go on Vacation in a few days and I wanted to "share" my progress/work, so I get Feedback that I can use when I'm back.


src/main/java/cam72cam/immersiverailroading/multiblock/SteamGeneratorMultiblock.java
```diff
@@ -0,0 +1,204 @@
...
+	private static final BlockPos output = new BlockPos(1, 0, -1);
+	
+	static MultiblockComponent[][][] structure = new MultiblockComponent[3][3][3];
+
+	static
+	{
+		for(int h = 0; h < 3; h++)
+			for(int l = 0; l < 3; l++)
+				for(int w = 0; w < 3; w++)
+					structure[h][l][w] = L_ENG();
```
EweLoHD: I forgot to change the MultiblockComponents, that were my test components xD

cam72cam:
This looks interesting.  I'll be hopping back on discord this weekend and will try to get through the backlog of PRs

src/main/java/cam72cam/immersiverailroading/multiblock/SteamGeneratorMultiblock.java
```diff
@@ -0,0 +1,230 @@
...
+			IFluidHandler water = getTile(watercap).getCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, null);
+			IFluidHandler steam = getTile(steamcap).getCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, null);
+			
+			player.sendMessage(new TextComponentString("Steam: " + String.valueOf(10000 - steam.fill(new FluidStack(IRFluids.FLUID_STEAM, 10000), false)) + " / 10000"));
+			player.sendMessage(new TextComponentString("Water: " + String.valueOf(10000 - water.fill(new FluidStack(FluidRegistry.WATER, 10000), false)) + " / 10000"));
+			//player.sendMessage(new TextComponentString("WaterHandler: " + " / " + waterHandler.toString()));
+			player.sendMessage(new TextComponentString("Origin: " + origin));
+			player.sendMessage(new TextComponentString("Water: " + getPos(watercap)));
+			player.sendMessage(new TextComponentString("Steam: " + getPos(steamcap)));
+			player.sendMessage(new TextComponentString("Power: " + getPos(power)));
```
EweLoHD: This was only for debugging. Needs to get removed

src/main/java/cam72cam/immersiverailroading/entity/LocomotiveSteam.java
```diff
@@ -664,8 +664,8 @@ public void onUpdate() {
 		} else {
 			if (boilerPressure > 0) {
 				// Reduce pressure by needed temperature
-				boilerPressure = Math.max(0, boilerPressure - (100 - boilerTemperature));
-				boilerTemperature = 100;
+				//boilerPressure = Math.max(0, boilerPressure - (100 - boilerTemperature));
```
cam72cam: Why comment this out?  This keeps some of the core (admittedly strange) mechanics working

EweLoHD: It's the only way that I found to get the Steam Loader working. When it's not commented out the Pressure gets always set to 0 if the Steam Loco isn't fired up. I didn't find any other fix for that.

cam72cam:
You need to save the tank states as part of the multiblock NBT still

Also, do we need a block for steam?  Does that make the rendering easier?
Do we want to make the multiblock bigger?  I don't know how big these should be IRL

src/main/java/cam72cam/immersiverailroading/entity/LocomotiveSteam.java
EweLoHD: That happens with those lines.

cam72cam: Ah, I see.  Maybe it's time to rewrite that code to be more realistic.

cam72cam: That code needs to condense steam which should increase the boiler heat till it hits 100

EweLoHD:
I removed the Steam Block because it's useless, but it changes nothing on the Rendering, the Texture still doesn't loads. 
Steam Generators in IRL can have all Sizes, so I think a 3x3 Multiblock is good.

EweLoHD:
I changed the Steam Condensing, but I'm not completely sure if it works as it should. 
I tested the Condensing and also the Augment and both worked.

EweLoHD:
So, I only need to get the Steam Texture to work before the pr could get merged. Does anybody know why the Texture doesn't load?

FriedrichLP:
I would suggest looking through the whole console (searching for your fluid block's name) and seeing what error there is. Minecraft should always inform you about texture errors

EweLoHD:
It doesn't have capital letters and there's nothing in the logs about missing steam textures, otherwise, it would be simple.

km00700:
how are you rendering steam, from the obj?

EweLoHD:
no. It's a normal Fluid, so I only need to load the Texture.

EweLoHD:
Wtf. When I reload my Textures in-game (F3+ T), the Texture is getting displayed. But when I restart my Game it's not getting displayed.

EweLoHD:
Registering Textures to the TextureMap separately should normally be not necessary, but it's the only fix for this. I will do another commit with removing all unused imports and then I think we should be able to merge the pr, or?

cam72cam:
I'll review this in detail later today and play around with it in game

src/main/java/cam72cam/immersiverailroading/multiblock/SteamGeneratorMultiblock.java
```diff
@@ -0,0 +1,204 @@
...
+				
+				//Output Steam
+				BlockPos outputPos = origin.add(output);
+				IFluidHandler output = FluidUtil.getFluidHandler(world, outputPos, EnumFacing.SOUTH);
+				if(output != null && steamFilled >= 10 && output.fill(new FluidStack(IRFluids.FLUID_STEAM, 10), false) >= 10) {
+					output.fill(steam.drain(10, true), true);
+					
+				}
+				
+			} catch (Exception e) {
```
cam72cam: Add an ImmersiveRailroading.catching here

src/main/java/cam72cam/immersiverailroading/multiblock/SteamGeneratorMultiblock.java
```diff
@@ -0,0 +1,204 @@
...
+				
+			}
+		}
+
+		@Override
+		public boolean canInsertItem(BlockPos offset, int slot, ItemStack stack) {
+			return true;
```
cam72cam: So any block in the structure can have items pumped in?  is that right?

src/main/java/cam72cam/immersiverailroading/tile/TileMultiblock.java
```diff
@@ -84,6 +90,35 @@ public int extractEnergy(int maxExtract, boolean simulate) {
     	}
     };
     
+    FluidTank tank = new FluidTank(10000);
+    
+    private IFluidHandler fluid = new IFluidHandler() {
```
cam72cam: Why define your own IFluidHandler?  FluidTank already implements IFluidHandler?

src/main/java/cam72cam/immersiverailroading/tile/TileRailBase.java
```diff
@@ -533,6 +540,10 @@ protected void onContentsChanged() {
 				}
 			};
 			break;
+		case STEAM_LOADER:
+			this.augmentTank = new FluidTank(1000);
```
cam72cam: Do you also need to filter this on Fluid_steam only?  Will it accept non fluid_steams after it's been emptied?

cam72cam:
steam_genertaor.obj is misspelled

src/main/java/cam72cam/immersiverailroading/multiblock/SteamGeneratorMultiblock.java
EweLoHD: no, I will fix that.

src/main/java/cam72cam/immersiverailroading/tile/TileMultiblock.java
EweLoHD: Otherwise, there is no FluidHandler attached to a Tile and then the Fluid Pipes (of IE) wouldn't connect.

src/main/java/cam72cam/immersiverailroading/tile/TileRailBase.java
EweLoHD: yes, because it's the only way (that I found) I can accept Steam Fluids form other Mods.

EweLoHD:
The (Creative) Capacitors of IE doesn't output Energy on their bottom Side.

cam72cam:
I don't think IR needs this block right now, given the required complexity.  It may be easier to get the interaction working right after the physics / heat overhaul that's coming sometime in the next year.  I think we can re-consider this when we do the multiblock / augment overhaul.

PR #435, opened 2018-09-23
support tool recipes if no steel ingots are available

FriedrichLP:


src/main/java/cam72cam/immersiverailroading/util/RecipeUtil.java
```diff
@@ -0,0 +1,13 @@
+package cam72cam.immersiverailroading.util;
+
+import cam72cam.immersiverailroading.ImmersiveRailroading;
+import net.minecraft.item.crafting.IRecipe;
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.registries.IForgeRegistryModifiable;
+
+public class RecipeUtil {
```
cam72cam: This should not need to be it's own class/function.  Please collapse this functionallity into registerRecipes

src/main/java/cam72cam/immersiverailroading/proxy/CommonProxy.java
```diff
@@ -150,6 +157,22 @@ public void serverStarting(FMLServerStartingEvent event) {
     
     public abstract World getWorld(int dimension);
     
+    @SubscribeEvent
+    public static void registerRecipes(RegistryEvent.Register<IRecipe> event) {
+    	IForgeRegistryModifiable<IRecipe> modRegistry = (IForgeRegistryModifiable<IRecipe>) event.getRegistry();
+    	if (!Loader.isModLoaded("immersiveengineering")) {
```
cam72cam: Don't check for IE.  Check that the ore dictionary contains an ingotSteel entry

PR #474, opened 2018-11-17
Fix for 22.5 angled switches not registering redstone

bplainia:
This fixes switches not picking up redstone in the proper spot by offsetting where the redstone detection origin is.
0 degree and 45 degree tracks will not be affected by the change.

Noted that there is a rendering error for switches facing left in any direction that is not equal to 0.


cam72cam:
Thanks!
What name should I use for you in the AUTHORS file?

bplainia:
> Thanks!
> What name should I use for you in the AUTHORS file?

You can do it like yours `Benjamin Plain (bplainia)`

PR #476, opened 2018-11-19
merge into master

FriedrichLP:


src/main/java/cam72cam/immersiverailroading/Config.java
```diff
@@ -53,6 +54,9 @@
 
 		@Comment("All gauges require fuel")
 		public static boolean FuelRequired = true;
+		
+		@Comment("Can players break trains")
+		public static TrainBreakMode trainBreakMode = TrainBreakMode.TRUE;
```
cam72cam: Just make this a boolean.  We don't need a state where ops can't break trains.

WrongWeekend:
Remember that players can still bypass this to an extent by using the wrench to quickly take apart and re-assemble stock elsewhere.

cam72cam:
Closing this as it should go into the locking changes

PR #485, opened 2018-11-24
Golden spike recipe

hayesj1:
After a lot of digging, I couldn't find where the recipes are actually loaded. If there's additional code to write please tell what it is.


cam72cam:
Looks good, thanks!  Can you put an entry in the Authors file?

hayesj1:
All set :)

cam72cam:
Thanks!

PR #493, opened 2018-12-02
Add info for each loco to OC detector.consist()

shotbyapony:
add information on each locomotive to OC component.ir_augment_detector.consist()

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/thirdparty/opencomputers/AugmentDriver.java
+++ b/src/main/java/cam72cam/immersiverailroading/thirdparty/opencomputers/AugmentDriver.java
@@ -1,8 +1,6 @@
 package cam72cam.immersiverailroading.thirdparty.opencomputers;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
 
 import cam72cam.immersiverailroading.Config;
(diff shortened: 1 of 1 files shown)


src/main/java/cam72cam/immersiverailroading/thirdparty/opencomputers/AugmentDriver.java
```diff
@@ -236,6 +237,44 @@ private FluidStack getFluid() {
 				}
 				info.put("direction", dir.toString());
 				
+				for (EntityCoupleableRollingStock car : stock.getTrain()) {
+					if (car instanceof Locomotive) {
+
+						EntityRollingStockDefinition def = car.getDefinition();
+
+						info.put("LOCO_"+locomotiveCount+"_id", def.defID);
```
cam72cam: Instead of copying and pasting this code from above, it should be broken out into a common function and added as a List of Dicts to the info dict.

cam72cam: Ex:
```
List<Object> locos = new ArrayList<Object>();
for (EntityCoupleableRollingStock car : stock.getTrain()) {
    if (car instanceof Locomotive) {
        locos.add(loco_info((Locomotive)car))
    }
}
info.put("locomotives", locos)
```

PR #495, opened 2018-12-09
Fix Wiki typos & Update Resourcepack list

hayesj1:
I wasn't able to test the links on my fork, but I doubled check the formatting on them.


cam72cam:
These files are generated from the Wiki, can you update that instead?

hayesj1:
Oops, I thought that _was_ the wiki. I'll update the _actual_ wiki then.

PR #505, opened 2018-12-17
Make physics more realistic.

CelestiaTheDryad:
I took a stroll through the physics code and changed many things around to more accurately model real life physics. This has a general effect of slowing down train acceleration. The changes are all local to the PhysicsAccumulator class, so I don't think it should be able to break anything else. It compiled and ran fine in my (admittedly limited) testing.


cam72cam:
This looks really good, I'll spin this up in my dev environment tomorrow and give it a whirl.

Could you add your info to the AUTHORS file as part of this PR?  Feel free to put as much or as little info there as you wish.

cam72cam:
In addition, the comments are clean and precise.  Great work!

src/main/java/cam72cam/immersiverailroading/physics/PhysicsAccummulator.java
```diff
@@ -35,55 +38,63 @@ public void accumulate(EntityRollingStock stock, Boolean direction) {
...
-		
-		// SHOULD THIS HAVE DIRECTION MULT?
-		double grade = -Math.tan(Math.toRadians(pos.rotationPitch % 90)) * Config.ConfigBalance.slopeMultiplier;
-		// lbs * 1%gradeResistance * grade multiplier
-		gradeForceNewtons += (stockMassLb / 100) * (grade * 100)  * 4.44822f;
+		rollingResistanceNewtons += 0.0015 * stockMassLb * poundToNewton;
+
+		// see http://hyperphysics.phy-astr.gsu.edu/hbase/mincl.html for slope physics
+		// negative because a positive rotationPitch means stock is pointing uphill
+		double gradeForceFactor = -Math.sin(Math.toRadians(pos.rotationPitch)) * Config.ConfigBalance.slopeMultiplier;
```
cam72cam: This seems to be doing weird things.  Can you try this out locally on a few different slopes?  It seems to be accelerating in weird ways

cam72cam: Actually, my settings are tweaked.  let me reset them and try again

cam72cam:
Ok, I am not entirely sure what's going on, but wheel slip is much more prominent.  I made a train of 100 55t hoppers headed by a big boy and tender and filled it with coal.  The code on master can just barely pull the load without slipping.  On this branch, I start slipping at 30 fully loaded cars.

Here's some code to auto fill any freight cars in the world (requires a relog after placement)

```diff
diff --git a/src/main/java/cam72cam/immersiverailroading/entity/CarFreight.java b/src/main/java/cam72cam/immersiverailroading/entity/CarFreight.java
index 5c3acb49..81e54ac5 100644
--- a/src/main/java/cam72cam/immersiverailroading/entity/CarFreight.java
+++ b/src/main/java/cam72cam/immersiverailroading/entity/CarFreight.java
@@ -1,6 +1,8 @@
 package cam72cam.immersiverailroading.entity;
 
 import cam72cam.immersiverailroading.registry.CarFreightDefinition;
+import net.minecraft.init.Items;
+import net.minecraft.item.ItemStack;
 import net.minecraft.world.World;
 
 public class CarFreight extends Freight {
@@ -27,4 +29,12 @@ public class CarFreight extends Freight {
        }
        
        //TODO filter inventory
+       protected void initContainerFilter() {
+               if (world.isRemote) {
+                       return;
+               }
+               for (int i = 0; i < cargoItems.getSlots(); i++) {
+                       this.cargoItems.setStackInSlot(i, new ItemStack(Items.COAL, 64));
+               }
+       }
 }
```

CelestiaTheDryad:
I'm struggling to find what you mean by wheel slip? Do you mean that the Big boy isn't powerful enough to pull the cars? Rolling resistance is the only thing that didn't change (as it was correct as far as I could tell). So since train acceleration power is down, it makes sense they wouldn't be able to pull as much.

I'll do some more testing and see if the numbers line up that way. I never tested trains too big to pull.

cam72cam:
It's weird, the locomotive did not stall, it just sits there and spins the wheels

CelestiaTheDryad:
https://github.com/cam72cam/ImmersiveRailroading/blob/a8cca6a5573c9acff2d81a9f0aaab92e2ee3a028/src/main/java/cam72cam/immersiverailroading/physics/PhysicsAccummulator.java#L86

https://github.com/cam72cam/ImmersiveRailroading/blob/a8cca6a5573c9acff2d81a9f0aaab92e2ee3a028/src/main/java/cam72cam/immersiverailroading/physics/MovementSimulator.java#L36

Bug was caused by the above checks. Trains that accelerated too slowly had their velocities clamped to zero. By my math, a Big Boy should be able to pull about 700 full 55T hoppers on new physics before getting stopped by rolling resistance. Fix is to reduce this threshold, I don't know if this minimum velocity change will affect any other code.

Kriegfox:
"I'm struggling to find what you mean by wheel slip?" Wheel slip will occur when there is not enough Normal force on the wheels (Meaning weight) to multiply against the coefficient of friction. It means bigboy isn't heavy enough to pull what you have placed behind it. If it wasn't powerful enough, it wouldn't spin, it just wouldn't move. When you break grip with the rail, it isn't because of a lack of power, its from an excess of power. Its easy to think that because the train is too heavy that the engine doesn't have the power, this is false. Wheel slip is the result of a lack of adhesion (normally attributed to not enough weight or improper weight distribution). A STALL is a lack of power.

CelestiaTheDryad:
Sure. It's actually 496. It think I forgot the base mass of the hopper the first time around.

Big Boy has a maximum tractive force of 602,178 N (from game tooltips). A fully loaded 55T hopper has a mass of 82600kg (25000kg base, plus 90 stacks * 64 blocks per stack + 10 kg per block).  The rolling resistance of a car is 0.0015 * gravity * mass of car. This is 1,214.22 N for the full 55T hopper. 602,178 / 1,214.22 = 496.

WrongWeekend:
Regarding loaded weight of the hoppers, the number of inventory spaces is still just an estimate, and not directly based on real specifications. Additionally cargo weight a bit of a gray area due to undefined weights of various objects in Minecraft (real or fictional).

Kriegfox:
rolling resistance also needs to take into account materials in contact(steel on steel), coefficient of friction, number of wheels and diameter of the wheels on the car, and rail conditions. Also rolling resistance does not take effect until the car is in motion (in reality), initial starting of a train should have to deal with static friction, which is around .45? for Steel on Steel. Rolling Resistance I think is .015, not .0015, if you have a different source of information I'd like to see it.

CelestiaTheDryad:
Rolling resistance: https://en.wikipedia.org/wiki/Rolling_resistance, coefficient examples section. 0.0015 was used before I got to this project and fits with what wikipedia says.

Static friction: As far as I can tell, Immersive Railroading has no simulation of static friction, in most cases it wouldn't make a difference to the way people use the mod.

Boxcar mass: Is approximated in the mod by 10 kg per block plus empty mass. See:
https://github.com/cam72cam/ImmersiveRailroading/blob/a8cca6a5573c9acff2d81a9f0aaab92e2ee3a028/src/main/java/cam72cam/immersiverailroading/entity/Freight.java#L217-L221

Kriegfox:
While I don't trust a wikipedia page. Even according to the page, Rolling Resistance is not a constant, infact it is a product of Normal Force * Coefficient of Rolling Friction.

Here's a good PDF [http://www.mhi.org/media/members/14220/130101690137732025.pdf](http://www.mhi.org/media/members/14220/130101690137732025.pdf)

Forged Steel on Steel floor gives a coefficient of Rolling Friction of 0.019, lets multiply that by 110,000 lbs = 2,090lbs of Rolling Resistance. Because F = CRF * N

F = Rolling Resistance (Or FORCE required)
CRF = Coefficient of Rolling Friction
N = Normal Force (Weight applied to the wheels)

I'm currently modeling a Consolidation Camelback from the 1880s. The current physics engine allows me to pull 14 loaded 55ton hoppers, with 55 US tons of blocks (at 10kg each) + a caboose, which rounds out to around 1100 US tons on a 65 block grade. Considering this grade is comparable to the grade the engine in real life had to endure and the load it would have been expected the carry, and did it at 11mph, I considered the current physics fairly accurate. However, I have been working on reworking the entire physics engine's equation set.

Kriegfox:
Also, you are correct, The mod currently doesn't take into account static friction (yet).

I have alot in the works in the background.

Kriegfox:
In the PDF, The equation: F = f * W/R is what the true rolling resistance equation is. 

Where:

F   =  the force required to overcome the rolling friction
f    =  the coecient of rolling friction (units must match same units as R (radius))
W  =  Load on the Wheel
R   =  Radius of the Wheel

This is the equation I plan to implement in the new physics code in the future.

CelestiaTheDryad:
The reason your coefficient is different is because you're working in Imperial units and you're using a different equation for rolling resistance. The coefficients on the wikipedia article are for the metric system and have the wheel radius calculation built in. Also, the current equation is literally coefficient * normal force so I don't know where you're getting a problem from. I never claimed it was constant.

I wasn't aware that rolling stock definitions had wheel radius. If so might as well use the more dynamic equation.

Kriegfox:
Currently they do not have a radius definition. (soon enough)

Equations don't change between Imperial and Metric unless the equation needs newtons instead of Mass inwhich case gravity is added as a multiplier. If the equation is written and calculated properly, the results should be the same.

I have sitting here infront of me, my personal copy of the 30th edition of The Machinery Handbook published in 2016. In here I have every equation of mechanics known to man. On page 159 on Rolling Friction it states:

Let:

W = Total Weight of rolling body or load on wheel in pounds (or newtons)
r = radius of wheel in inches (or centimeters)
f = coefficient of rolling resistance in inches (or centimeters)

THEN, resistance to rolling in pounds (or newtons) is given as: Fr = Wf/r

The equation works with Both Imperial and Metric interchangeably.

As an example the book lists iron on iron as having a Coefficient of rolling resistance of 0.02 inches (0.051cm) Which falls pretty close to my Forged Steel on Steel example of 0.019.

Kriegfox:
Granted, I am not used to working with the metric system. But there is no way, that equation should be coming up with almost 500 overloaded 55ton hoppers for a bigboy, something is way off.

CelestiaTheDryad:
The data I'm using for train cars is coming from the mod, not any real life source. So the Big Boy can currently pull 496 hoppers with my physics. The most likely sources of error would be the coefficient taken from wikipedia, or the data from the mod. I'm inclined to trust wikipedia, so if there's a data discrepancy I'd look at the mod's data first.

Also, keep in mind the 496 only includes rolling resistance. It doesn't take grades, air resistance, or static friction into account. Even if all the input data is accurate, no real life Big Boy would come close to hauling 496 hoppers.

Kriegfox:
Figured it out ( I think ), The reason the number is so high is because the number we're getting is the number of cars Big Boy could pull in theory at barely sustained motion if all 100% of tractive effort could be applied at that point. HOWEVER, Steam engines do not generate maximum tractive effort at low speeds like a diesel does. So, to compound the situation the number of cars drops dramatically even if you want to just sustain 10mph, due to needing increased force to accelerate, meanwhile the engine still isn't able to generate full power. Bigboy is rated at over 6200HP @ 41MPH. Its rated ability is at this speed. A steam engine's power curve starts low, then increases with speed, peaks, and drops off after rated speed.

CelestiaTheDryad:
That seams reasonable, a quick scan didn't find any simulation of that effect in the mod.

Kriegfox:
Indeed you won't find it. Like I said, I'm currently in the middle of rewriting all of the physics, there is a planned "advanced controls" setting for the mod (which is what I'm writing around). 

There are alot of planned additions to the physics engine, re-writes and tweaks. I've been working on diesel-electric traction recently which is a totally different ball game.

I have to head to bed, But, I imagine if Cam likes what youre up to, I'll see you around.

CelestiaTheDryad:
I've just noticed that all of the physics calculations are all based off of the one position used to init the physicsAccumulator (seems to always be the front of the train? or maybe its always a locomotive on the train?). This causes strange behavior, like a train on the side of a cliff remaining still because the "main" car is on flat track. I would fix this, except I have no idea how to get the current position of a piece of stock. I don't understand the manipulations of the positions list for the rolling stock, which is where I think the positions I want are stored.

The reference to pos in this line should be replaced by a reference to the stock's position, however that gets found.
https://github.com/CelestiaTheDryad/ImmersiveRailroading/blob/09041dd088ac4f1e502da40c388873969317a4aa/src/main/java/cam72cam/immersiverailroading/physics/PhysicsAccummulator.java#L48

FriedrichLP:
The position of the stock isnt seperately stored anywhere. Minecraft stores the position inside the entity. You can access that using Entity#getPositionVector()
Maybe you can hop onto discord so we can help you faster if you have any more questions

CelestiaTheDryad:
Any additional comments on these changes? It all seems to work great in my testing and there's been another IR release in the meantime that I'd thought this would be considered for.

cam72cam:
This is going to require a lot of testing.  For 1.5.0 I wanted to keep the physics as similar as possible as I was changing the track system out from under it.

After the 1.5.1 patch goes out we can start discussing merging some or all of these changes on to master.

CelestiaTheDryad:
Will redo pull request against modcore.

Issue #510, opened 2018-12-29
Add Force Switch item


hayesj1:
Yesterday someone asked for a item to force switches to toggle state when r-clicked. I'd like to implement it. Now that I've poked around the code, I think it's doable with a "forcedSwitch" field in RailInfo. The method, as far as I can tell, SwitchUtil.getSwtichState() seems to be the function which all other functions call to find out the state of a switch. This means, by adding a boolean in RailInfo, all that would be needed is to check it in the this function; if it's true, return SwtichState.TURN (no else clause). Also in BlockRailBase.onBlockActivated() there's a great spot to check if the item was this new item and toggle the boolean appropriately.

Thoughts? Suggestions? I don't know what to actually name the item, but I'm thinking something like "Switch Hammer"

hayesj1:
Another thought on the name/texture. I was also thinking it could be the stick part of a switch lever.

PR #511, opened 2018-12-30
Switch toggle item

hayesj1:
Implements #510 , but at the moment it doesn't work.


hayesj1:
I don't know how the first 4 commits got in there... Those are wiki typo-fixes. whoops.

hayesj1:
It works now!

PR #513, opened 2018-12-30
Switch toggle item

### LINKED ISSUE #510: Add Force Switch item

hayesj1:
Will close #510. There's still some polishing work left(see list below), but everything seems to be working.

- [x] Get an *official* texture/model (right it uses minecraft:stick)
- [x] Translate the item name (I'm monolingual sorry)
- [x] Add translation capability to tooltip and translate it, or just remove it
    - [x] Make the tooltip translatable and
    - [ ] translate it(I'm monolingual sorry)


src/main/resources/assets/immersiverailroading/wiki/en_us/casting-basin.txt
```diff
@@ -9,6 +9,7 @@ It is built from:
 * 126 x Sand
 *  86 x Blast Brick
 *  63 x Steel Scaffolding
+*   1 x Block of Steel
```
cam72cam: Please don't update the wiki files directly, I regenerate them from the github wiki before release

cam72cam:
In the future, could you try to avoid reorganizing imports in a PR that has other work in it?  It makes it harder to review

hayesj1:
Of course, I actually didn't know I hadn't already pushed those wiki changes. Is there an easy way to cherry pick commits when making a pull request?

src/main/resources/assets/immersiverailroading/wiki/en_us/casting-basin.txt
hayesj1: I right, now it's all coming back. I must have forgotten to delete those changes after I updated the github wiki. Sorry about that.

cam72cam:
It's fine for this PR, I'll work around it

hayesj1:
In the future I'll avoid using that option though.

src/main/java/cam72cam/immersiverailroading/proxy/CommonProxy.java
```diff
@@ -161,11 +133,13 @@ public static void registerRecipes(RegistryEvent.Register<IRecipe> event) {
     		modRegistry.remove(new ResourceLocation("immersiverailroading:hook"));
     		modRegistry.remove(new ResourceLocation("immersiverailroading:manual"));
     		modRegistry.remove(new ResourceLocation("immersiverailroading:track blueprint"));
+	      modRegistry.remove(new ResourceLocation("immersiverailroading:switch hammer"));
```
cam72cam: indentation

cam72cam:
What do you think about changing forced to a SwitchState enum?  None could be used as not set, use redstone switch value.  It would allow you to lock the switch in straight or curved regardless of redstone value.

hayesj1:
After thinking it over, I think I know what you mean. That wouldn't be too hard; r-click with a switch lever would just cycle through those choices?

Also do you think Switch lever is a good item name? And if so, should I leave the internal name as SwitchHammer or rename it to SwitchLever? (thinking ahead to actual switch levers/stands)

src/main/java/cam72cam/immersiverailroading/tile/TileRail.java
```diff
@@ -56,15 +50,15 @@ public double getMaxRenderDistanceSquared() {
 	}
 
 	public void setSwitchState(SwitchState state) {
-		if (state != info.switchState) {
-			info = new RailInfo(info.world, info.settings, info.placementInfo, info.customInfo, state, info.tablePos);
+		if (state != info.switchState && info.switchForced == SwitchState.NONE) {
```
cam72cam: I think we still want to set the switchState here, even if it is being overwritten by switchForced

cam72cam:
If you need help with the texture ping @modeler in discord.  I'd like to get this in by the weekend if possible.

src/main/java/cam72cam/immersiverailroading/tile/TileRail.java
hayesj1: Removed. But if the state passed to setSwitchState() isn't from SwitchUtil.getSwitchState(), there will be issues(aside from initializations of course).

hayesj1:
I just did a pull from master to incorporate the latest changes there. This should once again be conflict free.

hayesj1:
Now we are just waiting on the model files from Thomas I think

cam72cam:
@hayesj1 It seems to function quite well.  There are a few things needed before this can be merged:
* [ ] Fix the issue where leading / trailing trucks don't follow the forced switch state
* [ ] Add a message to send to the user about what state the track is now in when they hit the switch with the hammer
  * Switch Locked Straight
  * Switch Locked Turn
  * Switch Unlocked

I'll work on getting the textures working and on some of the rotation.

After that's done, we can merge this work

cam72cam:
Also:
```
Parsing error loading recipe immersiverailroading:switch hammer
com.google.gson.JsonSyntaxException: Unknown item 'immersiverailroading:item_switch_hammer'
``

**Comment by hayesj1** [comment]
Pretty sure the JSON error is fixed. I forgot to refactor the output item name when I did the switch hammer -> switch key refactor.

File: src/main/java/cam72cam/immersiverailroading/blocks/BlockRailBase.java [hunk 4]
```diff
@@ -294,6 +293,13 @@ public boolean onBlockActivated(World worldIn, BlockPos pos, IBlockState state,
 		Block block = Block.getBlockFromItem(stack.getItem());
 		TileRailBase te = TileRailBase.get(worldIn, pos);
 		if (te != null) {
+			if (stack.getItem() == IRItems.ITEM_SWITCH_KEY) {
+				if (!worldIn.isRemote) {
+					te.cycleSwitchForced();
```
cam72cam: Why not return the new state here so you don't need to look it up down below?

src/main/java/cam72cam/immersiverailroading/util/SwitchUtil.java
```diff
@@ -41,6 +41,18 @@ public static SwitchState getSwitchState(TileRail rail, Vec3d position) {
 			}
 		}
 
+		if (parent.info.switchForced != SwitchState.NONE) {
```
cam72cam: Maybe use isSwitchstateForced here?

cam72cam:
Please remove the wiki updates.  They are generated before release from the wiki itself

src/main/java/cam72cam/immersiverailroading/blocks/BlockRailBase.java
hayesj1: Great question. The answer is because I created the method before I knew I needed to retrieve the new state again. Just so I know we're on the same page, you mean have cycleSwitchForced() return the new state to use in the chat message calls?

src/main/java/cam72cam/immersiverailroading/util/SwitchUtil.java
hayesj1: Why not? I was actually considering removing that helper because it wasn't used, but that's a great use.

hayesj1:
Seriously!? They're still in the PR? Near the top it says I reverted them in early Jan....

hayesj1:
Nevermind, the reverts got reverted...

src/main/java/cam72cam/immersiverailroading/blocks/BlockRailBase.java
cam72cam: Correct

cam72cam:
Still need to fix the render

hayesj1:
I'm wondering if it's a desync between the parent TileRail and children TileRail....

hayesj1:
@cam72cam are all of the visible track blocks/te's instances of *RailGag? Meaning the parent *Rail blocks/te's aren't actually clickable. I guess another way to put it is: if I click on random track blocks, will I ever click on an instance of BlockRail, or will I always interact with a BlockRailGag?

cam72cam:
You can click on either

hayesj1:
so then what logic defines whether a given track block is a gag or not a gag? And is there only one non-gag per track segment(i.e. the set of track blocks placed by a single right-click of the Track Blueprint)?

cam72cam:
There's one non gag per track builder, but track builders can be nested

hayesj1:
RailInfo only exists for non-gag te's, correct? So that means a given builder has one railinfo instance, correct? And updates to RailInfo aren't pushed to sub-builders, right? So if b1 is nested in b2, changes to b2's info won't be propagated to b1? And switch builders have 3 builders(counting the realStraightBuilder), so all together that means if I'm not calling the correct parent, state stored in RailInfo might get desynced, right?

hayesj1:
Well after some more thorough testing, I think the bogey's are respecting the _redstone_ state, not the _switchState_ value.

hayesj1:
Found this in MovementSimulator. Could this cause the issue? `// Not using bogey yaw here, is that OK?
		Vec3d result = rail.getNextPosition(currentPosition, VecUtil.fromWrongYaw(distance, rotationYaw));`

hayesj1:
After some more digging, I'm not actually sure why the bogies aren't following the correct track sometimes. It seems to happen when the redstone state implies a different switch state than the forced state(ex. block is powered but switch is locked straight, vice versa). At this time, I believe the bug is buried somewhere in the `nextPosition()` logic, but after looking through that call chain I have no ideas. `SwitchBuilder.isOnStraight()` is another thought, but again nothing stands out to me as obviously wrong. Strangely, I can't find any calls that don't delegate to `SwitchUtil.getSwitchState()`, and that function should be fine as long as `SwitchBuilder.isOnStraight()` is fine.

cam72cam:
still bugged

hayesj1:
Sorry for my absence. I'll look into this again now, but if I remember correctly I needed more info on how the bogie-track-tracking logic works.

hayesj1:
After doing more thorough testing and making a table, I have determined that the bogies _always_ diverge, instead only when they should.

hayesj1:
@cam72cam is switchState client-side-only? Because the code where I cycle the switchForcedState is wrapped by a `!world.isRemote()` check

hayesj1:
That seems to have been the issue. Moving the `cycleSwitchForced()` call outside of the client-side check fixes the issue.

cam72cam:
hmm, https://github.com/cam72cam/ImmersiveRailroading/pull/513/files#diff-7f2f9a9e5ba45e5ebd1c34987993be4cR862 should be syncing it with the client

cam72cam:
Ah, might have found it:
tileSwitch.markDirty() needs to be called as that's the one you are modifying!

hayesj1:
> hmm, https://github.com/cam72cam/ImmersiveRailroading/pull/513/files#diff-7f2f9a9e5ba45e5ebd1c34987993be4cR862 should be syncing it with the client

How would I sync it with the client? And which line number(s) are in question?

cam72cam:
Your commits look good, taking a look now.

hayesj1:
If I'm reading the conflict right, you removed 4 or 5 imports from ClientProxy? I checked and they are still usages those classes in ClientProxy's code. Is that intentional?

cam72cam:
I might have cleaned up imports that are not used in master, but are still in use on your branch
I'd keep all the lines when you resolve the conflict and I can pair out uneeded imports later

hayesj1:
:+1:

cam72cam:
Works perfectly with:
```
--- a/src/main/java/cam72cam/immersiverailroading/tile/TileRailBase.java
+++ b/src/main/java/cam72cam/immersiverailroading/tile/TileRailBase.java
@@ -859,6 +859,8 @@ public class TileRailBase extends SyncdTileEntity implements ITrack, ITickable {
                        newForcedState = SwitchState.values()[( tileSwitch.info.switchForced.ordinal() + 1 ) % SwitchState.values().length];
                        tileSwitch.info = new RailInfo(world, tileSwitch.info.settings, tileSwitch.info.placementInfo, tileSwitch.info.customInfo, tileSwitch.info.switchState, newForcedState, tileSwitch.info.tablePos);
                        tileSwitch.markDirty();
+                       this.markDirty();
+                       this.getParentTile().markDirty();
                }
 
                return newForcedState;
```

cam72cam:
Once you add that, I'll merge it

hayesj1:
Is that what you had in mind?

PR #530, opened 2019-01-26
Added text feedback for paintbrush

km00700:
Stock now tells you when it has no variants so players know that the
brush is actually still working.


cam72cam:
Needs translation capability, but otherwise looks good

PR #531, opened 2019-01-26
Overhaul Train collisions with blocks

Endalion:
**Technically**
- All blocks inside the train bounding box are now collided with
- No collision for blocks at angles greater than 45 degrees (90 degree arc) from the forward/reverse direction in the XZ plane.
- Blocks with higher hardness require higher speeds to break; below which the block will immediately stop the train.
- Trains have new blockCollisionHardness (consist: blockCollisionForce), physics accumulator uses this to somewhat simulate the reaction from slamming into something.
- Blocks only contribute a reaction force in the direction toward the train, ie. reversing out of blocks is possible.
- An absolute speed limit, above which a collision will destroy the train
- Trains no longer break unbreakable blocks (Bedrock/Barrier, which have hardness = -1)

**Functionally**
- Driving through blocks now either breaks them and slows down the train, or stops the train entirely
- Trains ghosting through walls is severely reduced
- Buffer stops work (*hype!*) at low speeds, as expected
- Trains will explode when colliding above 60km/h, catastrophically above 100km/h

To do:
- [ ] Ensure the math applies for entire consist
- [ ] Make sure I am not making unoptimised functions
- [ ] Remove debug statements when ready to push


cam72cam:
Most of this looks good.  My main concerns are:

* Performance
* Integration with accumulate

Performance can be reasoned about and then tested using VisualVM and a large train.

Accumulate is a bit more interesting.  At 100KM/h, accumulate still runs about once per second or two.  At once per second, it is traveling ~28 meters.

I think the "proper" solution would be to use a look ahead for the current blockpos to determine the weight for each accumulate.  Take a look at getCouplerPosition for example, it takes a TickPos as an argument so it can calculate in the future.

Overall, I think this is definitely on the right track!

Endalion:
Would a look ahead involve collision checks at all TickPos in positions? I am also concerned with the performance implications of that. Also, getCouplerPosition in [definition?](https://github.com/cam72cam/ImmersiveRailroading/blob/master/src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java) doesn't have TickPos parameters

cam72cam:
https://github.com/cam72cam/ImmersiveRailroading/blob/master/src/main/java/cam72cam/immersiverailroading/entity/EntityCoupleableRollingStock.java#L672

I think we can find a way to optimize it

bigyihsuan:
I think because of these new block collision physics IR will need to add its own very hard block to act as a buffer. I don't think having a slab of obsidian will make sense in a modern-looking station building.

km00700:
IR already has plans for actual buffers, so it's not an efficient use of time to bodge something now.

Endalion:
> I think because of these new block collision physics IR will need to add its own very hard block to act as a buffer. I don't think having a slab of obsidian will make sense in a modern-looking station building.

Iron blocks, bars, anvils are able to stop trains up to 36km/h. I can imagine steel blocks and scaffold can do as well. I am not sure why you'd want a buffer with higher tolerance than this, but a long stretch of high-hardness blocks/longer stretch of lower hardness blocks will slow down and eventually stop the train over a distance.

Endalion:
changed to only call triggerResimulate when collision value changes (block broken, slam into wall, etc.)

cam72cam:
@Endalion you should be able to write it to do the block collision checks during triggerResimulate and store a list of block breakages inside of that (I think)

That's what I had in mind originally.

cam72cam:
I will try to review this soon, sorry it has taken me so long

PR #535, opened 2019-01-31
Safer dismounts

Endalion:
**Technically**
- Dismount position is independent of z-offset: anywhere in the right side of the cabin will dismount to the same distance from the right side of cab.
- Dismount will attempt to place the player up to 3m from the train, if all positions are obstructed it attempts default position (train's origin, offset to roof), else cancels dismount. Messages are provided to tell the player what went wrong.
- Standing still and holding down the dismount button will only check once*, player has to move from the last dismount attempt in the cabin to check again. *(twice apparently, I am unsure why.)

**Functionally**
- Players should never dismount into walls
- Players are able to dismount into tight spaces beside the train.

 **Issues**
- [x] Dismount distance may be too close
- [ ] Players cannot exit a completely covered train
- [ ] Holding down dismount button in cabin causes player to be rotation-locked
- [x] Default dismount position may not be safe at speed
- [ ] Performance?

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/entity/EntityRidableRollingStock.java
+++ b/src/main/java/cam72cam/immersiverailroading/entity/EntityRidableRollingStock.java
@@ -251,10 +251,9 @@ public abstract class EntityRidableRollingStock extends EntityBuildableRollingSt
 			
 			if (passenger instanceof EntityPlayer && !passenger.isDead && !this.isDead) {
 				if(!ppos.equals(passengerDismountAttempts.get(passenger.getPersistentID()))) {
-					delta = dismountFreePos(ppos, passenger);
+					delta = dismountFreePos(ppos, (EntityPlayer)passenger);
 					if (delta.y == -1) {
 						passengerDismountAttempts.put(passenger.getPersistentID(), ppos);
-						passenger.sendMessage(ChatText.DISMOUNT_FAIL.getMessage());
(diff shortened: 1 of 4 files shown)


src/main/java/cam72cam/immersiverailroading/entity/EntityRidableRollingStock.java
```diff
@@ -140,6 +145,7 @@ public boolean shouldRiderSit() {
 	}
 
 	public Map<UUID, Vec3d> passengerPositions = new HashMap<UUID, Vec3d>();
+	public Map<UUID, Vec3d> passengerDismountAttempts = new HashMap<UUID, Vec3d>();
```
cam72cam: This should probably not be a static variable

cam72cam:
This code should be integrated into dismountPos instead of having it's own logic.  Also, it smells pretty complex for something that should be a few simple position checks.

Endalion:
I didn't find this extra safety necessary for anything other than the player so I made it seperate from dismountPos.
I don't like the way I did it honestly, I was intending to check all possible overlapping blocks but I couldn't figure how to do it simply. Ended up snapping to the grid. There's also a lot of repeated code for checking blockpos.

src/main/java/cam72cam/immersiverailroading/entity/EntityRidableRollingStock.java
Endalion: It's static?

PR #539, opened 2019-02-08
Restrict building stock to only recommended and model gauges

Endalion:
Addresses [Config option for restricting stock to intended gauge](https://github.com/cam72cam/ImmersiveRailroading/issues/266)

"Model" gauges is defined by Gauge.isModel(gauge)


cam72cam:
Additionally, let's set this to "False" as a default for now.

Endalion:
New function to handle gauge cycle with filter
Button overlap on window resize fixed
Defaulted to false

PR #545, opened 2019-02-17
Render only the highest matching cargo load if showCurrentLoadOnly

Endalion:


FriedrichLP:
Your defaulting showCurrentLoadOnly to false is working. I must have forgot that. But the "fix" for the render will do exactly the same as my one except its more complicated.

Endalion:
This fix will skip all renders except the last one, the original condition did the exact opposite: It only rendered the first match, eg. ID_0 <= 100%, then immediately break.

PR #547, opened 2019-02-19
Sus horn fixed

EScapeTheShadows:
code allows for bells to play on locomotives when button is held down. also changes horn to being played only when key is held down, similar to new whistles


src/main/java/cam72cam/immersiverailroading/entity/LocomotiveDiesel.java
```diff
@@ -190,11 +204,12 @@ public void onUpdate() {
 					this.soundThrottle += Math.min(0.01f, absThrottle - this.soundThrottle);
 				}
 	
-				if (horn.isPlaying()) {
+				if (this.getDataManager().get(HORN) < 1) {
```
cam72cam: Leave this line alone.  The original code was clearer / standard

src/main/java/cam72cam/immersiverailroading/entity/Locomotive.java
```diff
@@ -25,7 +25,9 @@
 	private static DataParameter<Float> AIR_BRAKE = EntityDataManager.createKey(Locomotive.class, DataSerializers.FLOAT);
 	protected static DataParameter<Integer> HORN = EntityDataManager.createKey(Locomotive.class, DataSerializers.VARINT);
 	protected static DataParameter<Optional<UUID>> HORN_PLAYER = EntityDataManager.createKey(Locomotive.class, DataSerializers.OPTIONAL_UNIQUE_ID);
-	
+	protected static DataParameter<Integer> BELL = EntityDataManager.createKey(Locomotive.class, DataSerializers.VARINT);
+	protected static DataParameter<Optional<UUID>> BELL_PLAYER = EntityDataManager.createKey(Locomotive.class, DataSerializers.OPTIONAL_UNIQUE_ID);
```
cam72cam: Let's get rid of the BELL_PLAYER attribute.  We don't need to have the same person turn the bell on and off.

cam72cam:
Also need to fix the merge conflicts with master

src/main/java/cam72cam/immersiverailroading/entity/LocomotiveDiesel.java
EScapeTheShadows: i'll change it back, but for reference, this isn't a peice of the code i touched actually, so not sure how it got changed

src/main/java/cam72cam/immersiverailroading/entity/LocomotiveSteam.java
```diff
@@ -239,6 +243,13 @@ public void onUpdate() {
 					this.pressure = ImmersiveRailroading.proxy.newSound(this.getDefinition().pressure, true, 40, this.soundGauge());
 					pressure.setVolume(0.3f);
 				}
+
+				if (this.getDataManager().get(BELL) != 0 && !bell.isPlaying()) {
```
cam72cam: Please move the common bell code into Locomove.java

PR #550, opened 2019-02-23
Render changes (VBO optimizations)

cam72cam:
* Refactor out VBA code
* Modify VBA code to use a VBO subsystem if available
* Only upload a single copy of the model to the GPU
* Free model data (large memory savings) since it's all uploaded to the GPU anyways


cam72cam:
I want to get this on master so other people can start merging it into their PRs if there are conflicts

PR #553, opened 2019-02-24
Added Cab Car Type Locomotive

km00700:


cam72cam:
What's the difference between this and a diesel locomotive?

km00700:
no limitation on fuel, and as a separate type may be able to have control offered in only a certain part of the stock.

Kriegfox:
You understand a Cab car is by definition unpowered in most cases and controls a powered "locomotive" from the opposite end of the train?

km00700:
I'm well aware.

km00700:
Hence it doesn't require fuel, and will eventually only allow for a defined are to be controllable.

km00700:
I called it cab car locomotive due to the in-code class inheritance

PR #555, opened 2019-03-01
Custom Curves with Connecting Slopes

CelestiaTheDryad:
Custom curves now support having nonzero grades at either end.  Currently, the grade on each piece is selected by shift right clicking to up the grade by 1%, and shift left clicking to lower the grade by 1%. The current grade is outputted to the terminal.

This feature is not release ready, but it's gotten to areas in mod development I'm not at all familiar with and can probably be more effectively done by someone more familiar with IR. (Especially for the graphical issues).

- [x] Fix graphical glitches
- [x] Give proper feedback to player about grade selection
- [ ] Rigorous testing

Issues:
-There are some graphical glitches at the ends of slopes and where rails overlap.
-Floating point math sometimes causes flat tracks to destroy the blocks they are placed on and be very slightly sunk into the ground. (Most likely fixed by adding a small extra height to each rail)
-There's no feedback to the player what grade they have selected

Possible Expansion:
-Have the track blueprint/golden spike automatically take the grade from the piece of track they are used on. This is probably doable with onItemUse().


cam72cam:
Hmm, this is definitely interesting.  Figuring out a intuitive way to use this logic will be hard.

CelestiaTheDryad:
This feature is now ready for testing and review I believe.

CelestiaTheDryad:
will redo pull request against modcore.

PR #558, opened 2019-03-12
Bed placement fix

CelestiaTheDryad:
Fixes issues with rail beds not lining up correctly with rails.


cam72cam:
I'll try to get to this soon.  It fixes some of the weird math crap I've hacked.

I will need to set up a varried test world on master and then convert it over to test

PR #564, opened 2019-03-23
Indirect-Fire Railway Artillery

Endalion:
If you want to test this PR, use [this resource pack](https://www.dropbox.com/s/dvhzg1u1chktp0m/K5.zip). The model is modified from an internet source (License and Attribution present). I do not intend to include this stock in the base mod nor distribute it elsewhere, it is for testing and reference purposes.
In addition, it is recommended you have a waypoint indicator mod so you know where the big stick should point.

### Function ###
- Support for railway guns, on an external turret[1] or simple gimbal mount, indirect fire only
- Barrel + Breech + Turret[1] animation, Physical and Animated recoil impulse
- Target is chunkloaded; max potential range is World size. Minimum range likely present.
- Time of Flight, ie. the projectile will take time to "reach" the target
- GUI coordinate aiming, automatic gun slewing, chat feedback
- Gunpowder + Shell requirement. Shell from casting basin. (Creative players: gotta search)
- Rope trigger (Firing the gun by lead in hand, allows some use while not riding)

### Technical ### 
- Parabolic projectile trajectory (this isn't tangible anywhere except in variables), **air resistance is ignored**
- Dispersion is square, not circular.
- Projectile spawns at height limit and only acts (falls) in the terminal phase, therefore firing elevation is limited to above 45deg (where minimum range arises)
- Additional section in physics Accumulate for recoil force, though it acts more like an impulse than a force.
- Shell is loaded from the model. It's pretty hacky.
- Possible support for missiles. (Requires tweaks to aiming and animation)

### Issues ### 
- [ ] [1] Untested.
- [ ] Determine if chunkloading ammo requires restraint
- [ ] Projectile ignores blocks in its pre-terminal path (most notably in-front of the muzzle)
- [ ] Recoil force is not directional
- [ ] Recoil force may not be responsive in a high update-rate situation (throttle/brake is changing), untested.
- [ ] Consider model/effects for projectile, in-flight/terminal
- [ ] Go ahead with missile support or drop it

**The features above are not final and some might change.**


cam72cam:
@Endalion please remove the commit with the resource pack as I don't want stuff with different licenses in the IR repo.  Something like `git reset --hard HEAD~1` `git push -f`.  Stick the RP in dropbox or drive and leave the link in the description.

Endalion:
Resource pack removed

cam72cam:
@Endalion I gave it a try, it will need a better targeting mechanism, maybe raytrace from where the player is looking.  I can't seem to get it to aim anywhere.

Endalion:
Oh? There should be chat feedback when it can't aim; for reference, I targetted coords about 8 chunks away and it worked okay.
Targets too close aren't meant to be acquired since this is artillery and not full-on gun wagons. That would be a whole nother PR.

DrprofLuigi:
I made a railway gun as well, so there would be multiple packs if this gets implemented.

PR #573, opened 2019-04-10
Update Base Mod D&RGW Models

Thomas4peyton:
In progress...


cam72cam:
I'm going to close this for now as the plan is to move most of the base stock into a resource pack in the next year.  I will leave the branch around for when that happens

PR #577, opened 2019-04-12
add working reverser

FriedrichLP:


src/main/java/cam72cam/immersiverailroading/entity/LocomotiveSteam.java
```diff
@@ -162,6 +163,24 @@ protected int getAvailableHP() {
 		return (int) (this.getDefinition().getHorsePower(gauge) * Math.pow(this.getBoilerPressure() / this.getDefinition().getMaxPSI(gauge), 3));
 	}
 	
+	@Override
+	protected double getAppliedTractiveEffort(Speed speed) {
```
cam72cam: Find a way to implement this without copy/paste

src/main/java/cam72cam/immersiverailroading/entity/LocomotiveSteam.java
```diff
@@ -162,6 +163,24 @@ protected int getAvailableHP() {
...
+			reverser = Math.abs(0.8 - this.getReverser());
+			reverser = this.getReverser() > 0.0 ? reverser : -reverser;
+		}
+		
+		double outputHorsepower = Math.abs(Math.pow(getThrottle() * reverser, 3) * getAvailableHP());
+		
+		double tractiveEffortNewtons = (2650.0 * ((locoEfficiency * outputHorsepower) / Math.max(1.4, Math.abs(speed.metric()))));
+		return tractiveEffortNewtons;
+	}
+
```
Kriegfox: I'm going to write up the real equation for the reverser, locoefficieny should be something already a kfactor worked into output power by the model (which since 99% of them use SMEX that works a default Kfactor into the output answer). 

Also reading through this, while I'm not a coder, I can read what is getting multiplied and divided and I don't think this is going to have the desired result. I could be wrong. But let me write up the equation set for the reverser and you guys can compare it to what you have written.

cam72cam:
Closing in favor of https://github.com/TeamOpenIndustry/ImmersiveRailroading/pull/1032

PR #609, opened 2019-06-11
Tool to exchange tracks

FriedrichLP:


cam72cam:
Does this drop the current items and require new ones?

src/main/java/cam72cam/immersiverailroading/render/item/TrackExchangerModel.java
```diff
@@ -0,0 +1,188 @@
...
+		return new ItemOverrideListHack();
+	}
+
+	@Override
+	public Pair<? extends IBakedModel, Matrix4f> handlePerspective(TransformType cameraTransformType) {
+		Pair<? extends IBakedModel, Matrix4f> defaultVal = ForgeHooksClient.handlePerspective(this, cameraTransformType);
+		switch (cameraTransformType) {
+		case THIRD_PERSON_LEFT_HAND:
+		case THIRD_PERSON_RIGHT_HAND:
+			return Pair.of(defaultVal.getLeft(), new Matrix4().scale(1.5, 1.5, 1.5).translate(0.5, 0.5, 0.5).toMatrix4f());
```
cam72cam: Right/Left should not use the same settings, also what happens when the user clicks with the track in the off hand?

src/main/java/cam72cam/immersiverailroading/blocks/BlockRailBase.java
```diff
@@ -304,6 +318,34 @@ public boolean onBlockActivated(World worldIn, BlockPos pos, IBlockState state,
...
+				String track = ItemTrackExchanger.get(stack);
+				if (track != null && !track.equals(tileRail.info.settings.track)) {
+					if (!playerIn.isCreative()) {
+						RailInfo info = tileRail.info.withTrack(track);
+						if (info.build(playerIn, false)) { //cancel if player doesn't have all required items
+							FMLCommonHandler.instance().getMinecraftServerInstance().getPlayerList().sendPacketToAllPlayers( //we need to send the packet because this code is executed on the server side
+									new SPacketSoundEffect(SoundEvents.BLOCK_ANVIL_PLACE, SoundCategory.BLOCKS,pos.getX(), pos.getY(), pos.getZ(), 1.0f, 0.2f));
+							tileRail.info = info;
+							
+							List<ItemStack> drops = tileRail.getDrops();
```
cam72cam: This should call spawnDrops with a location vec3d parameter

src/main/java/cam72cam/immersiverailroading/tile/TileRail.java
```diff
@@ -130,6 +130,9 @@ public NBTTagCompound writeToNBT(NBTTagCompound nbt) {
 	public void setDrops(List<ItemStack> drops) {
 		this.drops = drops;
 	}
+	public List<ItemStack> getDrops () {
```
cam72cam: Remove this function and change spawnDrops to take a location (can add a spawnDrops function that takes no args and calls spawnDrops with this.pos).

PR #613, opened 2019-06-17
Allow stock tag get/set overrides

CSX8600:
There has been a reasonable request from RailStuff fans that they want to be able to return values set by the Rolling Stock Identifier using OpenComputers.  Since RailStuff does not depend on OpenComputers, RailStuff cannot rely on rolling stock tag as there's no way to set it without circumventing OpenComputers functionality.  These edits will allow RailStuff and other mods to hook into values set and returned by OpenComputers.


src/main/java/cam72cam/immersiverailroading/thirdparty/CommonAPI.java
```diff
@@ -157,10 +159,22 @@ public FluidStack getFluid() {
...
+    	
         return stock.tag;
     }
 
     public void setTag(String tag) {
+    	TagEvent tagEvent = new TagEvent(TagEvent.EventType.SetTag);
+    	tagEvent.tag = tag;
+    	MinecraftForge.EVENT_BUS.post(tagEvent);
+    	
         stock.tag = tag;
```
CSX8600: Still set stock tag even if another mod receives the event.  This way, calling setTag from OpenComputers will allow the tags to persist even if the mod that has event handlers are removed.

cam72cam:
How does railstuffs know what stock it's operating on?

CSX8600:
It uses a HashMap of the unique identifier of the Entity and the String identifier, but I think I see what you're getting at lol.  The UUID of the stock is never posted in the Event 🙃.  Good catch, will update after work!

cam72cam:
I'd create two different classes, both extending the TagEvent class, even if they are just extensions with no overrides.  This is a more "forge" way of doing this instead of with the enums.

I'd also make the tag a constructor param as it will be needed in both cases.

You could also simplify the stockID field to be public final and remove the getter.

cam72cam:
I'd also recommend you build a jar off of this branch and make sure that it works with railstuffs in the way you intend before merging this.

cam72cam:
This will also need to be rebased off of master to remove a few of Legend's misc commits

CSX8600:
> I'd create two different classes, both extending the TagEvent class, even if they are just extensions with no overrides. This is a more "forge" way of doing this instead of with the enums.
> 
> I'd also make the tag a constructor param as it will be needed in both cases.
> 
> You could also simplify the stockID field to be public final and remove the getter.

I can agree with you on everything but making the tag a constuctor param.  It is true that the SetTagEvent will supply listeners with a tag, but the GetTagEvent will be requesting a tag from listeners, therefore will always be null upon initialization.

cam72cam:
Ok, then only add the tag as a constructor param in the SetTagEvent.

CSX8600:
I believe I did the rebase successfully - let me know if I did not or if things still need changes yet.  I did test with RailStuff and everything worked flawlessly

src/main/java/cam72cam/immersiverailroading/thirdparty/event/GetTagEvent.java
```diff
@@ -0,0 +1,10 @@
+package cam72cam.immersiverailroading.thirdparty.event;
+
+import java.util.UUID;
+
+public class GetTagEvent extends TagEvent {
```
cam72cam: These classes can be moved to static classes inside of TagEvent, that's how forge does it a lot of the time.  Other than that, this PR looks good!

CSX8600: Done!

PR #622, opened 2019-07-14
New IConditionalCollision interface that lets blocks decide whether to allow collision

milandamen:
For a mod I'm making, I have a block that can be placed near a track. Normally the train would not hit this block, but if the train goes through a tight turn the block might be in the way of the train.

This PR adds the ability for blocks to skip collision checking with rolling stock so that those blocks do not get destroyed.

https://i.imgur.com/Iw9sRqF.png

I would love to hear your thoughts on this problem and this solution.

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/entity/EntityMoveableRollingStock.java
+++ b/src/main/java/cam72cam/immersiverailroading/entity/EntityMoveableRollingStock.java
@@ -1,25 +1,19 @@
 package cam72cam.immersiverailroading.entity;
 
-import java.util.ArrayList;
-import java.util.List;
-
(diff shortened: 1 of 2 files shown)


cam72cam:
I'll probably be moving the interface location so if you are using this in another mod, it will probably need to be updated when I merge the modcore branch in.

cam72cam:
@milandamen Let me know what info you'd like me to put in the AUTHORS file!

milandamen:
@cam72cam Thank you very much! I agree with moving the interface to a different directory, I just didn't know what directory it should have been in when I made it.

`Milan Damen (milandamen)` in authors is fine with me.

PR #650, opened 2019-09-18
Refactor IR on top of a Modcore

cam72cam:
This will allow Immersive Railroading to maintain 99.9% of the same code between minecraft versions.  The "cam72cam.mod" package is what will be changed between versions.

I don't plan to have this perfectly bug-free before merge, but I'd like the majority of functionality working with minimal crashes.

TODO:
* Consider moving modcore to it's own repo and add a subrepo link / package a prebuilt jar
* Bugfixes:
    *  [ ] Look at crashes inside multiblock init code (could have been from weird test setup at the time)
    * [ ] Verify multiblock functionality (power/item input/etc...)
    * [ ] Add default texture (grey) while icons/main textures are loading
    * [ ] Remove render debug output
    * [ ] Consider not moving a train until all cars are loaded / decoupling after certain period of time (will probably be a different PR)
    * [x] Consider pulling knowledge of TrackAPI into modcore (would remove remaining net.minecraft usage / hackery)

I'll also be looking at a 1.7.10 and 1.14 port to test how well the modcore abstraction works.  I won't require 100% functionallity, just the basics to prove the concept.

Modcore will see a good number of changes/bugfixes over time as it was a massive rewrite which introduced a lot of bugs and pulled some bad behavior through from old code.  Over time it will stabilize.  In the future it may be a standalone mod which can load other jars.  This would allow the same IR jar to be used in multiple MC versions / mod loaders!  I've done some ugly PoC work with tricking forge into accepting fake mods on the fly so it is possible.

I appreciate all of the developers waiting a long time on this and hope that we can get back to a more healthy environment where I can respond to PRs and actually merge them in at a decent cadence.


cam72cam:
@milandamen I've moved where that ICollidableThing interface is in the code, you will need to make sure you update when I release the next IR version.

milandamen:
@cam72cam Thanks for notifying me. It looks good to me.

The only thing I noticed in the `EntityMoveableRollingStock` code change is that the `intersects` check gets called before the `IConditionalCollision` check gets called. The order doesn't really matter, but in the old code there was a comment that said that the `intersects` call is pretty slow. Maybe it would make sense to switch the `intersects` and `IConditionalCollision` checks around so `intersects` gets checked after.

For blocks that don't implement IConditionalCollision it would not matter, but it would be a small improvement in speed when a block does implement IConditionalCollision and returns false on that check.

Thanks for all the hard work on this mod!

cam72cam:
This is not yet complete, but I am splitting the repo so I need to get this in.

PR #658, opened 2019-10-05
Enable getName function for Common APIs

Simba98:
- The Immersive Railroading Integration should use this function. Instead of call name() function directly.


cam72cam:
This getName function is to deal with the minecraft entity name.  It was disabled due to how I am doing entities in modcore and will be reenabled later on.

The CommonAPI for computers already uses the definition name which is going to be more correct anyways: https://github.com/TeamOpenIndustry/ImmersiveRailroadingIntegration/blob/forge_1.12.2/src/main/java/cam72cam/immersiverailroading/thirdparty/CommonAPI.java#L59

PR #666, opened 2019-10-17
New physics

CelestiaTheDryad:
See #505

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,5 +17,6 @@ Jacob Hayes (tigerbird1)
 Mason Besmer (HealthCareUSA)
 Milan Damen (milandamen)
 Kuzmin V...w (Starter)
+Brendan Thomas (CelestiaTheDryad/Ultra695)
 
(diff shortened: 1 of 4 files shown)


cam72cam:
I'll give this a try tonight.  If it is more accurate than what's currently in there (which it probably is) I'll merge it.  I know @Kriegfox is working on perfection, but that can be built on top of this work.

JustDesoroxxx:
So what was the result of the trial?

cam72cam:
The underlying coupling/netcode/"physics" is currently all tied together in a bad way.  My next project after the render overhaul is to refactor that (late spring / summer project).  My goal is to clean up the code so other folks can add/modify physics without worrying about the crazy set of interdependent pieces.

This stuff grew organically over a few years and I'm putting in the work to reign it in / fix it up so other folks can more easily contribute to it.

Desoroxxx:
Why did this close as it been implemented?

PR #718, opened 2020-01-02
Added multithreading to stock model loading and generating height maps.

milandamen:
This PR adds multithreading to the stock model loading and height map generation process to make it go faster.

Previous duration: 8 seconds
PR duration: 4 seconds
(Tested with 32 cores, but should go fast on any multi-core system)

I have looked into making sprite rendering go faster, but I don't think it's possible to multithread that.


andrboot:
Could you possible do caching as well ? :P

milandamen:
@andrboot I'm not sure what you mean by caching? The stock definitions get loaded at game startup and then don't get unloaded until you close the game. When you do anything with stock it doesn't re-load the information, it just uses the information already in memory.

Unless you mean inbetween game starts?

andrboot:
Yeah, so it doesn't need to generate icons/heightmaps unless they change 😬

FriedrichLP:
A lot of caching is going to be done through RenderLib. It already caches models and their textures so they don't have to be parsed again which makes the whole process about 25 times faster.

andrboot:
@FriedrichLP was thinking more on startup, so multi-loads doesn't need to re-render

andrboot:
@milandamen here is a nice crashlog!
[crash-2020-01-03_19.25.10-client.txt](https://github.com/TeamOpenIndustry/ImmersiveRailroading/files/4018496/crash-2020-01-03_19.25.10-client.txt)

milandamen:
@andrboot Seems one or more of the models could not be loaded for you. It does work correctly for me, but I have improved the exception message.

Could you send me the whole log, because that will contain the actual error.

andrboot:
Crashed out earlier this time :)
 See logs.
[crash-2020-01-04_07.37.21-client.txt](https://github.com/TeamOpenIndustry/ImmersiveRailroading/files/4020818/crash-2020-01-04_07.37.21-client.txt)
[latest.log](https://github.com/TeamOpenIndustry/ImmersiveRailroading/files/4020820/latest.log)
[debug.log](https://github.com/TeamOpenIndustry/ImmersiveRailroading/files/4020821/debug.log)

milandamen:
@andrboot Weird.. Could you try deleting the `build` directory in the ImmersiveRailroading directory and building again? If this does not fix it, could you send me the assets/immersiverailroading/rolling_stock/stock.json file inside the ImmersiveRailroading .jar? (Provided you don't override this file with a resource pack).

I have made a commit that will log which stock could not be loaded.

andrboot:
Each build = clean build environment in Jenkins :) 
Basically.. have all the stock publicly available  :P

cam72cam:
Have you considered using a parallel stream with a lambda?  That would make this code much cleaner and not require all the boiler plate / separate classes.

milandamen:
Thanks for the pointer towards parallel streams. I have refactored the code to use them and it is much easier to read now. The performance is also very similar in comparison to using a completion service.

src/main/java/cam72cam/immersiverailroading/registry/DefinitionManager.java
```diff
@@ -78,98 +79,174 @@ public static void initDefinitions() throws IOException {
...
+                    bar.step(stockDefinition.name());
+                    definitions.put(stockDefinition.defID, stockDefinition);
+                }
+            } catch (Exception e) {
+                ImmersiveRailroading.error("Error loading model %s of type %s", defID, defType);
+                ImmersiveRailroading.catching(e);
+
+                // Important so that progress bar steps correctly.
+                synchronized (monitor) {
+                    bar.step("");
```
cam72cam: Could be synchronized(bar), does not require the extra object

src/main/java/cam72cam/immersiverailroading/registry/DefinitionManager.java
```diff
@@ -78,98 +79,174 @@ public static void initDefinitions() throws IOException {
...
-        List<InputStream> inputs = blacklist_json.getResourceStreamAll();
+    public static Set<String> getDefinitionNames() {
+        return definitions.keySet();
+    }
+
+    public static Collection<TrackDefinition> getTracks() {
+        return tracks.values();
+    }
+
+    public static List<String> getTrackIDs() {
```
cam72cam: Could you move these functions back down in the file so they don't show up in the diff?

cam72cam:
Looks pretty good.  I'll take a deeper look tonight and try to spin it up.  @andrboot could you try this branch with your setup?

andrboot:
Initial load works much better, no crash this time on boot, nicely maxed out cores.
Will test more this evening with your latest commit @milandamen

andrboot:
Testing worked last night, no errors apart from stock errors which looks like invalid json.
No abnormal crashes :)

cam72cam:
@milandamen this looks pretty good!

I have one last question before merge, what happens if someone has a high number of CPUs and has not allocated enough memory to MC?

I had a similar situation in the texture loading (it does 2 at a time IIRC) because it would get stuck doing a few large textures in parallel and spike the memory usage.

A) We could determine this experimentally by artificially restricting XmX both with and without these changes.

B) If it is a problem, we could limit the max number of parallel streams as a factor of expected memory per job.

What are your thoughts?

milandamen:
In theory the parallel streams could indeed use more memory than when single-threaded, because loading multiple stock at once would use `(mem per stock load) * threads + (time until garbage collection)` amount of memory.

I've just read a bit about how parallel streams divide tasks over threads, and it seems to always be `(num of processors) - 1`, unless you use a global override which would affect all parallel streams. You could also use a custom `ForkJoinPool` with a custom number of threads, but there is no guarantee that this will work in all JVMs, because it is apparently an implementation detail and not in the spec.

https://stackoverflow.com/questions/30802463/how-many-threads-are-spawned-in-parallelstream-in-java-8

I think this is a disadvantage of using parallel streams compared to a thread pool with completion service.

In my opinion using parallel streams is still better than thread pool with completion service, because of the readability of the code, so I think we have a few possible options:

1. Have the user set `-Djava.util.concurrent.ForkJoinPool.common.parallelism=8` (replace 8 with num of processors), but this would affect all parallel streams, and would depend on JVM implementation.
2. Use a custom `ForkJoinPool` with a custom number of threads (based on IR config), but would increase code complexity, and would depend on JVM implementation.
3. Add an IR config option that lets the user disable multithreaded loading of stock, but would increase code complexity by a little (extra `if` statement that gets type of stream based on config), and would be a hard all threads or 1 thread choice.

Of course the best option the user can take is increase XmX but these options would be relevant if that is not possible.

I would love to hear your opinion on this.

cam72cam:
I think a config option is a good place to start.  Additionally, we could look at free memory and divide by num CPUs -1.  If that is below say 50MB, we could disable the parallel stream call and log a message.

milandamen:
What should I call the config option, and in which file?

Maybe a new "performance" category in the "general" category of immersiverailroading.cfg with a config name like "multithreadedStockLoading" with default `true`?

I think a check for free memory is going to muddy the code too much, at least for now.

cam72cam:
The config you proposed sounds good.  I might move some existing options there.

Checking free memory should be as simple as https://stackoverflow.com/a/25596.

You should be able to create a function that takes a stream <t> and returns a stream<t> that has optionally had parallelStream called (based on config and memory/CPU).

If we don't do that I think this would need to be disabled by default.

cam72cam:
If you want, we could merge as is and I could take care of this check

milandamen:
I was doing some testing and I did confirm that when low on memory, multithreaded does indeed get stuck while singlethreaded continues. I did some other testing and I think the sweet spot might be around 1264 MiB free memory (after manual garbage collection to get the true free memory), though the testing can probably be done better.

Test results: https://gist.github.com/milandamen/e2fedd651166cdd0c25c62af2ae7e37f

Sadly I didn't have much time this week to work on it more.
I can create some code for this, this weekend, and then we can check if it works like it should.

cam72cam:
No rush, just wanted to make sure I was not holding it up.  Based on those numbers, I think 50 MB/CPU makes sense

milandamen:
No problem :)

Yeah something around 50 MiB/CPU should be fine. (I have 32 cores total, so this load is divided over 31 cores)

milandamen:
I've made a commit that adds the memory checking and config option.

I've tried it with various memory sizes and enabling/disabling the config option and it seems to work. For my system (32 core, so stream uses 31 cores), it uses singlethreaded stream till 1.5GB memory, then from 2GB memory onward it uses multithreaded stream. When disabling the config it always uses singlethreaded stream even on high memory setting (12GB).

src/main/java/cam72cam/immersiverailroading/registry/DefinitionManager.java
```diff
@@ -234,6 +242,41 @@ private static JsonObject getJsonData(String defID) throws IOException {
...
+
+        // Parallel streams use numCPUs-1 threads for stream workloads.
+        Runtime runtime = Runtime.getRuntime();
+        int processors = runtime.availableProcessors() - 1;
+        if (processors <= 1) {
+            return collection.stream();
+        }
+
+        // Manual garbage collection so we get an accurate quantity of free memory.
+        runtime.gc();
```
cam72cam: Excellent idea

src/main/java/cam72cam/immersiverailroading/registry/DefinitionManager.java
```diff
@@ -234,6 +242,41 @@ private static JsonObject getJsonData(String defID) throws IOException {
...
+
+        long maxMemory = runtime.maxMemory();
+        long totalMemory = runtime.totalMemory();
+        if (maxMemory == Long.MAX_VALUE) {
+            maxMemory = totalMemory;
+        }
+
+        long freeMemory = runtime.freeMemory();
+        freeMemory += maxMemory - totalMemory;
```
cam72cam: I don't understand this part?  Would this double the amount of detected free memory?

cam72cam:
I've got that one last question, and then it's ready to merge.  Can you add an entry to the Authors file?  Add as much or as little information as you want.

src/main/java/cam72cam/immersiverailroading/registry/DefinitionManager.java
milandamen: So for example when you have an initial heap size (Xms) of 1GB and a maximum heap size (Xmx) of 2GB, when this code gets the amount of free memory, it can be from a total anywhere between 1GB and 2GB. When more memory is needed the heap size gets increased until it hits 2GB.

So to get the true amount of free memory you need to know how big the heap currently is (totalMemory) and to what size it can grow (maxMemory). You then add the size difference to the indicated freeMemory, because when growing the heap, that extra memory becomes available to use.

milandamen:
I'm already in the authors file, but thanks for the reminder. I have also answered your question about the free memory calculation, and I hope I explained it well enough. I based the calculation on the stackoverflow page I linked earlier in this PR and after I read some of the documentation.

src/main/java/cam72cam/immersiverailroading/registry/DefinitionManager.java
cam72cam: Ah, thanks for the explanation!

cam72cam:
Thanks!

PR #786, opened 2020-04-23
Remove empty tag compound so items stack

littlej541:
The steel ingot and steel block cast out of the casting basin do not stack with the same items acquired elsewhere because an empty tag compound is applied to the item. This clears the tag compound if it is empty right before placing the item into the casting basin output title entity container.

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/multiblock/CastingMultiblock.java
+++ b/src/main/java/cam72cam/immersiverailroading/multiblock/CastingMultiblock.java
@@ -252,7 +252,11 @@ public class CastingMultiblock extends Multiblock {
 					if (mode == CraftingMachineMode.SINGLE) {
 						craftTe.setCraftMode(CraftingMachineMode.STOPPED);
 					}
-					outTe.getContainer().set(0, item.copy());
+					ItemStack outputItem = item.copy();
+					if (outputItem.getTagCompound().internal.hasNoTags()) {
+						outputItem.clearTagCompound();
+					}
+					outTe.getContainer().set(0, outputItem);
 				} else {
 					if (craftTe.getRenderTicks() % 10 == 0) {
 						if (fluidTe.getCraftProgress() > 0) {


FriedrichLP:
This is a good change but as far as I know this current change does not work. You cannot access ModCore internal stuff from IR. The hasNoTags() has to be added to ModCore and then it can be used in here

littlej541:
I know it works with what I forked from master and built, but I wouldn't put it past myself to have done something wrong with the build. Unless you're talking about some upcoming major change or something.

FriedrichLP:
I'm sorry for being a bit unclear. This code definitely works in 1.12.2 and technically is fine on this version. The issue is that this might/will break in a different version of IR. With UMC all the code from IR remains the same between every version. To make this work cam spent half a year abstracting away the minecraft code into UMC. So in order to make fully working code you will have to use the methods inside the TagCompound and not it's internal. Bascially as soon as you use the "internal" variable inside any UMC class you will almost definitely break some version that IR supports. Since hasNoTags() does not yet exist in TagCompound it will have to be added (through a PR to the UMC repository).
(UMC=UniversalModCore=modcore)

littlej541:
Got it. Then would something like `outputItem.getTagCompound().toString().equals("{}")` work? Or is there something else I'm missing something there too. That UMC stuff looked tough.

FriedrichLP:
That could work but it wouldn't be too easy for others to understand (harder to understand than a simple hasNoTags()). You could probably leave the PR in the current state then. I can add the mentioned method to ModCore soon and then you can update the PR to use that method instead. After that I think this PR is ready to be merged.

littlej541:
Sorry, I didn't mean to imply toString() would be a good alternative. That's why I did hasNoTags() rather than the above mentioned, I just didn't realize the significance of internal since I could access its members. I was just wondering if toString() would work because I thought it was the one from `Object`, but I now see it's overridden in ItemStack in UMC. It looks simple enough in ItemStack though. I just remember seeing a bunch of lambdas, generics, and annotations somewhere. In any case, I'll wait for the update to UMC then.

PR #977, opened 2021-05-14
Stock reorganization

km00700:
**Reorganized Stock Hierarchy:**

- Added ControllableStock interlayer to handle control inputs and MU logic. (any controllable stock can be configured to operate in MU consist)
- Added fuel-free locomotive type for locomotive types with external fuel sources. (handcar, cable-driven, electric)

**Updated GUI Overlays:**

- Renamed GUI overlays to generalized names. (more universal applications)
- Added brake-only GUI background image.

**Added Stock Types:**

- Added Control Cars to operate a MU-capable locomotive from the end of a consist.
- Added Brake Vans to apply extra braking from a specific car.
- Added Cable-driven locomotive type for cable cars and incline railways


Appw0:
Will this allow for stocks without brakes? In an implementation of brake vans I did on my fork a little ago, I added a tag for braking force to all stock, thus allowing brake vans and also brakeless cars.

km00700:
Not as-is. That would have to be done separately by overriding the current system. The idea of this system was to keep it similar to the existing system, but abstract the logical layers. I can add a flag for brakeless stock later this evening.

src/main/java/cam72cam/immersiverailroading/registry/LocomotiveDieselDefinition.java
```diff
@@ -16,14 +16,15 @@
...
-    public boolean muliUnitCapable;
     private FluidQuantity fuelCapacity;
     private int fuelEfficiency;
     private boolean hornSus;
 
     public LocomotiveDieselDefinition(String defID, JsonObject data) throws Exception {
         super(LocomotiveDiesel.class, defID, data);
 
+        //diesels allow MU by default
+        multiUnitCapable = true;
```
cam72cam, line 27: This ignores the setting further in the file.

cam72cam:
Closing as I ended up implementing most this a different way

PR #980, opened 2021-05-15
Use items while on train

Appw0:
A hacky fix to allow instant use items, such as OC tablets, to be used while riding a train. This really shouldn't be how its done, but I suppose its step in the right direction that can be built upon before being merged.

Uses changes to UMC in https://github.com/Appw0/UniversalModCore/commit/ccc04c96f8d0b389e68ebf67feddb488afad79dd

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/entity/Freight.java
+++ b/src/main/java/cam72cam/immersiverailroading/entity/Freight.java
@@ -106,9 +106,11 @@ public abstract class Freight extends EntityCoupleableRollingStock {
 			}
 		}
 		
-		if (guiType() != null) {
+		if (guiType() != null && player.getHeldItem(Player.Hand.PRIMARY).isEmpty() && player.getHeldItem(Player.Hand.SECONDARY).isEmpty()) {
 			guiType().open(player, this);
 			return ClickResult.ACCEPTED;
+		} else if (!player.getHeldItem(Player.Hand.PRIMARY).isEmpty()) {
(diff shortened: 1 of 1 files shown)


cam72cam:
Honestly, I'm not sure there's a good alternative here.  AFAIK there's not a good spot to handle this in MC's netcode

Appw0:
How do horses and other mounts work? You can shoot a bow while on one. Maybe something can be found there. I looked into it but I don't have much experience with the minecraft source and haven't found it yet.

cam72cam:
That's a good question.  I think some of the "large entity clicking" stuff I wrote is breaking it.

cam72cam:
https://github.com/TeamOpenIndustry/UniversalModCore/blob/1.12.2-forge/src/main/java/cam72cam/mod/input/Mouse.java#L28-L37 is where it's breaking

Appw0:
Now all interactions ( lever pulling, block breaking ) can happen while riding a car that's hitbox is below the player, such as a flatcar, and item use ( bow shooting, book opening ) can now happen on all stock. This fits all the criteria I was looking for in these changes, and I'm not sure what else to do now. UMC commit is here https://github.com/Appw0/UniversalModCore/commit/5c26d76217d696bb3f72207700ca77c2be07aeab

cam72cam:
Sorry this took me a while @Appw0, I ended up doing a hybrid approach where the packet is only fired if ACCEPTED is returned.

PR #1041, opened 2021-08-13
Lighting support

cam72cam:


S4MillionYT:
Do i need any other mod than immersive railroading to get lights

PR #1097, opened 2021-10-10
Control automation fixes

cam72cam:


src/main/java/cam72cam/immersiverailroading/library/LocoControlMode.java
```diff
@@ -3,13 +3,11 @@
...
 
 public enum LocoControlMode {
-	THROTTLE_FORWARD,
-	THROTTLE_REVERSE,
+	THROTTLE,
+	REVERSER,
 	BRAKE,
 	HORN,
 	COMPUTER,
-	REVERSER_FORWARD,
```
cam72cam, line 11: Need dummy so test build / patreon world builds don't break

PR #1102, opened 2021-10-13
add extra tooltip infos

hayesj1:
Add the ability for pack makers to add extra tooltips to their models via `stock.json`s, using the new field `extra_tooltip_info`; this is an array of strings, where each element is one line of tooltip.


src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java
```diff
@@ -543,6 +549,7 @@ public String name() {
         tips.add(GuiText.WEIGHT_TOOLTIP.toString(this.getWeight(gauge)));
         tips.add(GuiText.MODELER_TOOLTIP.toString(modelerName));
         tips.add(GuiText.PACK_TOOLTIP.toString(packName));
+        tips.addAll(extraTooltipInfo);
```
cam72cam, line 552: NullPointerException!

PR #1295, opened 2022-11-13
feat: add config to reset train speed [blocked]

123FLO321:
- adds a debug config option to reset speed when loading stock
- this is useful when automating trains (avoid incorrect switch while during cc computers are still booting)


cam72cam:
This conflicts with https://github.com/TeamOpenIndustry/ImmersiveRailroading/pull/1242.  We can revisit adding a startup delay / zeroing speed after that is wrapped up and merged.  Hopefully that can happen before December.

123FLO321:
replaced by #1361

PR #1296, opened 2022-11-13
feat: add config for default augment modes

123FLO321:
- add debug option to automatically switch newly placed augments into computer mode


src/main/java/cam72cam/immersiverailroading/tile/TileRailBase.java
```diff
@@ -109,6 +109,11 @@ public void setAugment(Augment augment) {
 		this.augment = augment;
 		if (getParentTile() != null) {
 			augmentGauge = getParentTile().info.settings.gauge;
+			if (augment == Augment.DETECTOR && ConfigDebug.defaultAugmentComputer) {
```
cam72cam, line 112: I think you want an outer if statement for `ConfigDebug.defaultAugmentComputer`, with the existing if else inside of that.

cam72cam (changes requested):
Looks pretty straightforward, just a small note on the logic

cam72cam (approved):

PR #1307, opened 2022-12-18
Paintbrush rework

cam72cam:
Demo: https://youtu.be/N6exjzsbjqQ


src/main/java/cam72cam/immersiverailroading/entity/EntityRollingStock.java
```diff
@@ -143,6 +144,11 @@ private ClickResult selectNewTexture(Player player, ItemStack item) {
 		}
 	}
 
+	public void setTexture(String variant) {
+		// TODO packet
```
hayesj1, line 148: looks like you forgot to sed a packet :P

src/main/java/cam72cam/immersiverailroading/gui/PaintBrushPicker.java
```diff
@@ -0,0 +1,207 @@
...
+            return true;
+        });
+        this.updateVariants("");
+
+        width = 60;
+        Button random = new Button(screen, GUIHelpers.getScreenWidth() / 2 - width, ytop, width, height, "Random") {
+            @Override
+            public void onClick(Player.Hand hand) {
+                List<String> choices = new ArrayList<>(stock.getDefinition().textureNames.keySet());
+                variant = choices.get(r.nextInt(choices.size()));
```
hayesj1, line 92: There's a regression here. It's possible to get the same variant twice in a row, which is both a confusing UX and an annoyance for stock with only a few liveries.

PR #1308, opened 2022-12-18
Diesel Startup Sound

bakinser:
PartSound now has support for a start clip and (tentatively awaiting implementation) an end clip. New sound tag in json "start" which will play before the main engine loop upon starting a locomotive if present. I decided against making the start sound a default because even if you could find a good CC-free sound, it wouldn't sound good mixed with other pack's custom engine sounds, so it will be something modelers have to add themselves.

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/model/part/PartSound.java
+++ b/src/main/java/cam72cam/immersiverailroading/model/part/PartSound.java
@@ -71,13 +71,12 @@ public class PartSound {
                     startSound.setVolume(volume);
                     startSound.setPitch(pitch);
 
-                    if(!startSound.isPlaying() && !startHasPlayed) { //If nothing has started yet
-                        startSound.play(stock.getPosition());
-                        startIsPlaying = true;
-                    } else if (!startSound.isPlaying() && startIsPlaying) { //Start sound is done playing one take
+                    if (!startSound.isPlaying() && startIsPlaying) { //Start sound is done playing one take
                         startHasPlayed = true;
                         startIsPlaying = false;
-                        startSound.stop();
+                    } else if(!startSound.isPlaying() && !startHasPlayed) { //If nothing has started yet
+                        startSound.play(stock.getPosition());
+                        startIsPlaying = true;
                     } else {
                         startSound.update();
                     }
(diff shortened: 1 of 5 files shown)


src/main/java/cam72cam/immersiverailroading/model/part/PartSound.java
```diff
@@ -9,16 +9,34 @@
...
 public class PartSound {
-    private final Function<EntityMoveableRollingStock, ISound> create;
+    private final Function<EntityMoveableRollingStock, ISound> loopCreate;
+    private final Function<EntityMoveableRollingStock, ISound> startCreate;
+    private final Function<EntityMoveableRollingStock, ISound> endCreate;
 
-    public PartSound(Function<EntityMoveableRollingStock, ISound> create) {
-        this.create = create;
+    public PartSound(Function<EntityMoveableRollingStock, ISound> loopCreate)
+    { this.startCreate = null; this.loopCreate = loopCreate; this.endCreate = null;}
```
cam72cam, line 17: This can be much cleaner by calling other constructors.  Ex for this line:
```
this(null, loopCreate, null);
```

src/main/java/cam72cam/immersiverailroading/model/part/PartSound.java
```diff
@@ -31,36 +49,73 @@ public void effects(EntityMoveableRollingStock stock, float volume) {
...
 
     public void removed(EntityMoveableRollingStock stock) {
-        ISound sound = sounds.get(stock.getUUID());
-        if (sound != null) {
-            sound.terminate();
+        Sounds soundTrio = sounds.get(stock.getUUID());
+        if (soundTrio != null) {
+            if(soundTrio.start != null){ soundTrio.start.terminate();}
+            if(soundTrio.loop != null){ soundTrio.loop.terminate();}
+            if(soundTrio.end != null){ soundTrio.end.terminate();}
```
cam72cam, line 118: sounds.remove(stock.getUUID())

cam72cam:
A  few small comments, but overall looks good :+1:

src/main/java/cam72cam/immersiverailroading/model/part/PartSound.java
cam72cam: Took a second look and this is not something you introduced, but is a bug in the original code.

cam72cam: Thanks for pointing out that it's an ExpireableMap and that's not needed!

PR #1313, opened 2022-12-25
Add build instructions

Insprill:
I had a little bit of difficulty figuring out how to set up the development environment for IR, so this PR adds some instructions to assist people with that. If there are any mistakes or additional information that should be noted, please let me know. I used a footnote for linking an example of how to run UMCSetup, if you'd like me to remove that or display the example another way please let me know. This also fixes a small typo.


cam72cam:
Thanks!

PR #1323, opened 2023-01-09
Remove serializing into final java fields

cam72cam:
My evil wizard powers are not effective in newer java versions...


cam72cam:
This is going to need some serious validation

cam72cam:
Nordale has been running this branch without issue, I'll keep poking at it before release

PR #1332, opened 2023-01-17
Fix CurseForge README badges

Insprill:
[The docs](https://support.curseforge.com/en/support/solutions/articles/9000206928-curseforge-badges) say you can use the project slug, but it appears that no longer works.


cam72cam:
Thanks!

PR #1338, opened 2023-01-29
Caml support

cam72cam:
TODO:
- [x] Specifications
- [x] Testing/Verification
- [x] Code review

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/gui/overlay/GuiBuilder.java
+++ b/src/main/java/cam72cam/immersiverailroading/gui/overlay/GuiBuilder.java
@@ -52,18 +52,18 @@ public class GuiBuilder {
 
     protected GuiBuilder(DataBlock data) throws IOException {
         // common stuff
-        this.x = data.getValue("x").getFloat(0f);
-        this.y = data.getValue("y").getFloat(0f);
+        this.x = data.getValue("x").asFloat(0f);
+        this.y = data.getValue("y").asFloat(0f);
         DataBlock centered = data.getBlock("centered");
         if (centered != null) {
-            this.centerx = centered.getValue("x").getBoolean(false);
-            this.centery = centered.getValue("y").getBoolean(false);
+            this.centerx = centered.getValue("x").asBoolean(false);
+            this.centery = centered.getValue("y").asBoolean(false);
         } else {
             this.centerx = this.centery = false;
         }
 
         // Image stuff
-        this.image = data.getValue("image").getIdentifier(null);
+        this.image = data.getValue("image").asIdentifier(null);
         if (image != null) {
             BufferedImage tmp = ImageIO.read(this.image.getResourceStream());
             imageWidth = tmp.getWidth();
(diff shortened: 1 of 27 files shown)


src/main/java/cam72cam/immersiverailroading/gui/overlay/GuiBuilder.java
```diff
@@ -1,23 +1,23 @@
 package cam72cam.immersiverailroading.gui.overlay;
 
-import cam72cam.immersiverailroading.entity.*;
+import cam72cam.immersiverailroading.entity.EntityRollingStock;
```
hayesj1, line 3: `*` imports are annoying. Good change.

src/main/java/cam72cam/immersiverailroading/gui/overlay/GuiBuilder.java
```diff
@@ -74,79 +74,86 @@ protected GuiBuilder(JsonObject data) throws IOException {
...
-        this.control = data.has("control") ? data.get("control").getAsString() : null;
-        this.invert = data.has("invert") && data.get("invert").getAsBoolean();
-        this.hide = data.has("hide") && data.get("hide").getAsBoolean();
-
-        if (data.has("translate")) {
-            JsonObject tl = data.get("translate").getAsJsonObject();
-            this.tlx = tl.has("x") ? tl.get("x").getAsFloat() : 0;
-            this.tly = tl.has("y") ? tl.get("y").getAsFloat() : 0;
+        String readout = data.getValue("readout").asString();
+        this.readout = readout != null ? Readouts.valueOf(readout.toUpperCase(Locale.ROOT)) : null;
```
hayesj1, line 88: could this cause issues on different JVMs? Shouldn't there just be an "official" IR Locale used for data exchange between the mod and the models?

src/main/java/cam72cam/immersiverailroading/util/CAML.java
```diff
@@ -0,0 +1,172 @@
...
+
+                lines.remove(0);
+
+                String trimmed = val.trim();
+                if (trimmed.startsWith("\"") && trimmed.endsWith("\"")) {
+                    val = val.substring(1, val.length()-1);
+                }
+
+                if (StringUtils.isWhitespace(val)) {
+                    Block block = new Block(lines);
```
hayesj1, line 92: yikes, a recursive ctor?

src/main/java/cam72cam/immersiverailroading/util/DataBlock.java
```diff
@@ -0,0 +1,207 @@
...
+            @Override
+            public Double asDouble() {
+                return null;
+            }
+
+            @Override
+            public String asString() {
+                return null;
+            }
+        });
```
hayesj1, line 53: This is a fair bit strange in Java. I'd recommend having a static "emptyValue" sentinel that's reused, rather than using an anonymous class like this just for null-safety.

src/main/java/cam72cam/immersiverailroading/util/CAML.java
```diff
@@ -0,0 +1,172 @@
...
+        return new Block(lines);
+    }
+
+    private static class Block implements DataBlock {
+        private final Map<String, Value> primitives = new LinkedHashMap<>();
+        private final Map<String, List<Value>> primitiveSets = new LinkedHashMap<>();
+        private final Map<String, DataBlock> blocks = new LinkedHashMap<>();
+        private final Map<String, List<DataBlock>> blockSets = new LinkedHashMap<>();
+
+        private Block(List<String> lines) throws IOException {
```
hayesj1, line 54: These propbably shouldn't be IOExceptions, I think there's a Std Lib exception for parsing errors.

src/main/java/cam72cam/immersiverailroading/gui/overlay/GuiBuilder.java
cam72cam: I'm not super familiar with java locales, I just defaulted to this as it seemed like a sane default recommended by others.

src/main/java/cam72cam/immersiverailroading/util/CAML.java
cam72cam: Yep, I could break out parsing / constructing.  This was the simplest path at the time.  It's not terrible IMO.

src/main/java/cam72cam/immersiverailroading/util/DataBlock.java
cam72cam: Yeah, I'll probably move all the JSON related stuff into a helper class where I can do that instead of hacking it into an interface.

src/main/java/cam72cam/immersiverailroading/util/CAML.java
cam72cam: Fair point

src/main/java/cam72cam/immersiverailroading/registry/DefinitionManager.java
```diff
@@ -265,30 +293,50 @@ private static void initTracks() throws IOException {
...
         List<InputStream> inputs = track_json.getResourceStreamAll();
         for (InputStream input : inputs) {
+            blocks.add(JSON.parse(input));
+        }
+
+        Identifier track_caml = new Identifier(ImmersiveRailroading.MODID, "track/track.caml");
+        inputs = track_caml.getResourceStreamAll();
+        for (InputStream input : inputs) {
+            blocks.add(CAML.parse(input));
+        }
```
hayesj1, line 309: This seems to be a very common pattern. I'd make a utility method for it.

PR #1361, opened 2023-04-07
feat: add debug stick to set trains home points

123FLO321:
Add a new Item that is only obtainable in the creative inventory to set trains home locations.
When a location is set the train is teleported there when loaded.
This is useful for automated setups to recover when a server crashes.


cam72cam:
I don't think this is the right approach.  When a server crashes, the best solution is applying a backup.

PR #1362, opened 2023-04-07
fix: fix and refactor train overhead detection

123FLO321:
- the old logic for resetting `overhead` incorrectly always set the overhead back to null
- with this logic there is no extra collision calculation necessary


cam72cam:
This inspired a similar and simpler approach taken here: 45f30347

Issue #1368, opened 2023-04-09
Physics Performance Issues


123FLO321:
Currently the physics engine takes up a lot of the servers tick time.
On my server, for example with 3 trains running `cam72cam.immersiverailroading.entity.physics.SimulationState.next()` takes up `41%` of the total server main threads time.
78% of ticktime within `SimulationState.next()` is taken up by `cam72cam.immersiverailroading.tile.TileRailBase.getNextPosition()`.

`getNextPosition` has a complex while loop that is executed every phyics tick.
This could potentially be re-written to only be calculate once on loading/placing and then cache next positions for given headings and switch states.

![image](https://user-images.githubusercontent.com/1949038/230782020-98af90af-ec60-4824-9d1b-4937018dc481.png)
[sampler-snapshot.zip](https://github.com/TeamOpenIndustry/ImmersiveRailroading/files/11185766/sampler-snapshot.zip)

cam72cam:
A few questions:
Which build?
How many pieces of stock and at what speed?

cam72cam:
getNextPosition is complicated by a few concerns:
1) Overlapping pieces of track
2) Switches (similar to 1)

Depending on location and direction, the output vector can be dramatically different.

123FLO321:
It's build from #1362.
Event with only 1 Train (1 Locomotive, 3 Wagons) going at 200km/h the server tick time increases from 30ms to 50ms.
In the snapshot had an additional train running with 2 Locos, 8 wagons at. 160km/h.

Do you have any ideas how to improve this?
One possibility might be to calculate and save centre paths in the rail tiles (with branches for switches).
Then it would be possible to give the path a distance and get a position without much calculation.
Then the path only needs to be calculated once or when connected nodes change.

Storing such a path might also allow for potential built in automation in the future.

cam72cam:
The paths are already cached, it's figuring out where along the path / which path that takes the time.

200km/h in minecraft requires a significant amount of computation, not just by IR.  While it is supported, moving at that speed is definitely tricky.

cam72cam:
I've recently merged in changes that dramatically reduce the overhead of moving trains on a server.  I am going to close this issue.

PR #1369, opened 2023-04-09
perf: physic performance improvements

123FLO321:
- replace some distanceTo calls with distanceToSquared where possible for slight physics performance increase

(related to #1368)

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/track/IIterableTrack.java
+++ b/src/main/java/cam72cam/immersiverailroading/track/IIterableTrack.java
@@ -12,17 +12,17 @@ public interface IIterableTrack {
     List<BuilderBase> getSubBuilders();
 
     default double offsetFromTrack(RailInfo info, Vec3i pos, Vec3d position) {
-        double dist = 100;
+        double distSquared = 10000;
 
         // Convert to relative
         position = position.subtract(info.placementInfo.placementPosition).subtract(pos);
 
         for (Vec3d gagPos : getPath(info.settings.gauge.scale()/8)) {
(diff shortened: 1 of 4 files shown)


src/main/java/cam72cam/immersiverailroading/track/IIterableTrack.java
```diff
@@ -12,17 +12,17 @@ public interface IIterableTrack {
     List<BuilderBase> getSubBuilders();
 
     default double offsetFromTrack(RailInfo info, Vec3i pos, Vec3d position) {
-        double dist = 100;
+        double distSquared = 10000;
```
cam72cam, line 15: 100 * 100 would be a bit clearer and be optimized out by the compiler

src/main/java/cam72cam/immersiverailroading/track/IIterableTrack.java
```diff
@@ -12,17 +12,17 @@ public interface IIterableTrack {
...
-            double off = gagPos.distanceTo(position.add(0, -(position.y % 1), 0));
-            if (off < dist) {
-                dist = off;
+            double offSquared = gagPos.distanceToSquared(position.add(0, -(position.y % 1), 0));
+            if (offSquared < distSquared) {
+                distSquared = offSquared;
             }
         }
-        return dist;
+        return distSquared;
```
cam72cam, line 26: Math.sqrt(...)

cam72cam:
This looks really good, other than a few minor comments.

You can also improve TileRailBase: `double distanceMeters = motion.length();`

PR #1372, opened 2023-04-15
feat: add option to run physics on a separated thread [blocked]

123FLO321:
This PR adds an option to offload physics caclulation to a separated thread.
This makes the impact IR has on Server TPS next to 0.


123FLO321:
closes #1368

src/main/java/cam72cam/immersiverailroading/util/PhysicsThread.java
```diff
@@ -0,0 +1,66 @@
...
+import cam72cam.immersiverailroading.ImmersiveRailroading;
+
+import javax.annotation.Nullable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * Utility class for the physics thread.
+ */
+public class PhysicsThread {
```
cam72cam, line 13: Let's move this to the physics package

src/main/java/cam72cam/immersiverailroading/ImmersiveRailroading.java
```diff
@@ -110,6 +111,8 @@ public void commonEvent(ModEvent event) {
 				MultiblockRegistry.register(CastingMultiblock.NAME, new CastingMultiblock());
 				IRFuzzy.applyFallbacks();
 				break;
+			case START:
+				PhysicsThread.start();
```
cam72cam, line 115: This logic should be moved into PhysicsThread.run.  It can check if the executor is started in there.

123FLO321:
refactored implementation based on your suggestion and added handling for multiple dimensions

src/main/java/cam72cam/immersiverailroading/physics/PhysicsThread.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Gets or creates the physics thread, if enabled, or null if disabled.
+     * @return The physics thread, or null if disabled.
+     */
+    @Nullable
+    private static ExecutorService getExecutor() {
+        if (ConfigPerformance.multithreadedPhysics) {
+            if (executor == null) {
+                ImmersiveRailroading.info("Starting physics thread");
+                executor = Executors.newSingleThreadExecutor((t) -> new Thread(t, "ImmersiveRailroading Physics Thread"));
```
cam72cam, line 37: This could now could have multiple threads

src/main/java/cam72cam/immersiverailroading/entity/physics/Simulation.java
```diff
@@ -17,32 +18,66 @@ public class Simulation {
...
+     * Automatically dispatches physics simulation to a separate thread if enabled.
+     * @param world The world to simulate.
+     */
     public static void simulate(World world) {
-        // 100KM/h ~= 28m/s which means non-loaded stationary stock may be phased through at that speed
-        // I'm OK with that for now
-        // We might want to chunk-load ahead of the train just to be safe?
+        long tick = world.getTicks();
+        ServerChronoState chrono = (ServerChronoState) ChronoState.getState(world);
+        List<EntityCoupleableRollingStock> allStock = world.getEntities(EntityCoupleableRollingStock.class);
```
cam72cam, line 29: Could move the getEntities call into the if statement for efficiency

src/main/java/cam72cam/immersiverailroading/physics/PhysicsThread.java
cam72cam: Maybe that could be ConfigPerformance.multithreadedPhysics, where any value greater than 0 is the number of threads to use

cam72cam:
I've found some pretty dramatic performance improvements in the pathing code.  I'm going to close this pull request for now as the performance even on large servers is much less of a concern.

PR #1398, opened 2023-09-07
Change detection of when trucks/bogies need custom pathing

cam72cam:


cam72cam:
There will likely be followup work for specifying bogey/locomotive rotation points

PR #1465, opened 2024-03-24
Make left-first crank pin an adjustable feature

Goldenfield192:
In the past IR makes valve gears to be at left-first crank pin position by default,while many stocks are right-first.
These code adds "left_first" for stocks to change the crank pin's default position.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,4 +37,3 @@ jobs:
     - uses: actions/upload-artifact@v2
       with:
         name: Immersive Railroading ${{matrix.branch}}
-        path: build/libs/ImmersiveRailroading-${{matrix.branch}}*
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,3 +37,4 @@ jobs:
     - uses: actions/upload-artifact@v2
       with:
         name: Immersive Railroading ${{matrix.branch}}
+        path: build/libs/ImmersiveRailroading-${{matrix.branch}}*
--- a/src/main/java/cam72cam/immersiverailroading/model/part/DrivingAssembly.java
+++ b/src/main/java/cam72cam/immersiverailroading/model/part/DrivingAssembly.java
@@ -39,25 +39,14 @@ public class DrivingAssembly {
             return null;
         }
 
-        ValveGear left,
-                inner_left,
-                center,
-                inner_right,
-                right;
+        int angleDirection = isLeftFirst ? 1 : -1;
 
-        if(isLeftFirst) {
-            left = ValveGear.get(wheels, type, provider, state, ModelPosition.LEFT.and(pos), 0);
-            inner_left = ValveGear.get(wheels, type, provider, state, ModelPosition.INNER_LEFT.and(pos), 180);
-            center = ValveGear.get(wheels, type, provider, state, ModelPosition.CENTER.and(pos), -120);
-            inner_right = ValveGear.get(wheels, type, provider, state, ModelPosition.INNER_RIGHT.and(pos), 90);
-            right = ValveGear.get(wheels, type, provider, state, ModelPosition.RIGHT.and(pos), center == null ? -90 : -240);
-        }else {
-            left = ValveGear.get(wheels, type, provider, state, ModelPosition.LEFT.and(pos), 0);
-            inner_left = ValveGear.get(wheels, type, provider, state, ModelPosition.INNER_LEFT.and(pos), -180);
-            center = ValveGear.get(wheels, type, provider, state, ModelPosition.CENTER.and(pos), 120);
-            inner_right = ValveGear.get(wheels, type, provider, state, ModelPosition.INNER_RIGHT.and(pos), -90);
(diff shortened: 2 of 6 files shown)


Goldenfield192:
I mis-reseted the branch so there are 2 commits :(

SteampsonJake:
Kinda irrelevant now tbh since you can have fully custom valve gear with the Animatrix

Larky2k:
Yup pretty pointless now with Animatrix

Goldenfield192:
> Kinda irrelevant now tbh since you can have fully custom valve gear with the Animatrix

That's certainly true but turning it into an adjustable settig will also certainly save some resource pack authors a lot of time

Larky2k:
How are you saving time, if your not modelling using animatrix everything must be I full back position so im not seeing how your saving time?

cam72cam:
I think it's valid, some folks may not want to use Animatrix.

src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java
```diff
@@ -470,6 +471,9 @@ public void loadData(DataBlock data) throws Exception {
         DataBlock properties = data.getBlock("properties");
         weight = properties.getValue("weight_kg").asInteger() * internal_inv_scale;
         valveGear = ValveGearConfig.get(properties, "valve_gear");
+        if(valveGear !=  null){
+            isLeftFirst = properties.getValue("left_first").asBoolean(true);
```
cam72cam, line 475: This will need to be added to the default caml files.

src/main/java/cam72cam/immersiverailroading/model/part/DrivingAssembly.java
```diff
@@ -35,12 +39,25 @@ public static DrivingAssembly get(ValveGearConfig type, ComponentProvider provid
...
-        ValveGear inner_right = ValveGear.get(wheels, type, provider, state, ModelPosition.INNER_RIGHT.and(pos), 90);
-        ValveGear right = ValveGear.get(wheels, type, provider, state, ModelPosition.RIGHT.and(pos), center == null ? -90 : -240);
+        ValveGear left,
+                inner_left,
+                center,
+                inner_right,
+                right;
 
+        if(isLeftFirst) {
+            left = ValveGear.get(wheels, type, provider, state, ModelPosition.LEFT.and(pos), 0);
```
cam72cam, line 49: Instead of copying and pasting this code, let's define a "direction offset" and multiply the angle by that.
```java
int angleDirection = isLeftFirst ? 1 : -1
...
right = ValveGear.get(wheels, type, provider, state, ModelPosition.RIGHT.and(pos), angleDirection * (center == null ? -90 : -240));
```

cam72cam (changes requested):
I think this is a reasonable change, but there needs to be a few changes before this can be merged.

src/main/java/cam72cam/immersiverailroading/model/part/DrivingAssembly.java
Goldenfield192: Ok and I will change it later
Thanks

.github/workflows/release.yml
```diff
@@ -37,4 +37,3 @@ jobs:
     - uses: actions/upload-artifact@v2
       with:
         name: Immersive Railroading ${{matrix.branch}}
-        path: build/libs/ImmersiveRailroading-${{matrix.branch}}*
```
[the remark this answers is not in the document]
Goldenfield192: > Why was this removed?

I rolled back wrongly :(

src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java
cam72cam, line 475: Instead of using .asBoolean(true), lets uncomment the value in default/base.caml.  I've tried to move to that system so the documentation and code don't get out of sync.

Goldenfield192: > Instead of using .asBoolean(true), lets uncomment the value in default/base.caml. I've tried to move to that system so the documentation and code don't get out of sync.

Fixed it

cam72cam:
I'll give this a spin after work.  Would you be interested in adding yourself to the AUTHORS file as part of this PR?

Goldenfield192:
> I'll give this a spin after work. Would you be interested in adding yourself to the AUTHORS file as part of this PR?

That's my pleasure

src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java
Goldenfield192: > Instead of using .asBoolean(true), lets uncomment the value in default/base.caml. I've tried to move to that system so the documentation and code don't get out of sync.

But sir won't this lead to some capability issues?

cam72cam:
Hi!  Apologizes for not getting to this last week, my schedule has been crazy.

Goldenfield192:
> Hi! Apologizes for not getting to this last week, my schedule has been crazy.

Oh sorry for it and I'll make a new pr(this is way too ugly

PR #1470, opened 2024-03-31
Re: Make left-first crank pin an adjustable feature

Goldenfield192:
In the past IR makes valve gears to be at left-first crank pin position by default,while many stocks are right-first.
These code adds "left_first" for stocks to change the crank pin's default position.
This new pr has 1 commits instead of 7

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,3 +37,4 @@ jobs:
     - uses: actions/upload-artifact@v2
       with:
         name: Immersive Railroading ${{matrix.branch}}
+        path: build/libs/ImmersiveRailroading-${{matrix.branch}}*


src/main/resources/assets/immersiverailroading/rolling_stock/default/base.caml
```diff
@@ -48,6 +48,7 @@ properties =                # Required:
 #       -1.0 = immersiverailroading:some/path/animation_-1.0.anim # Reverse
 #       0.0 = immersiverailroading:some/path/animation_0.0.anim # Neutral
 #       1.0 = immersiverailroading:some/path/animation_1.0.anim # Forward
+    left_first = true            # Required if valve_gear is defined: Should this valve gear be at left-first crank pin position? Set false to make it right-first
```
cam72cam, line 51: What are your thoughts on changing this to a suffix on the valve_gear property?

valve_gear = WALSCHAERTS_INVERT

That's pretty easy to detect in ValveGearConfig.  This flag is also not applicable for the animatrix valve gear type, which this change would make clear.

Goldenfield192: That's a fantastic proposal. I'll try it

PR #1473, opened 2024-04-07
Changes based on current valve gear type

Goldenfield192:
Make IR's walschaerts valve gear can be used at right-first crank pin position
Thanks to @cam72cam


cam72cam:
I think you misunderstood my suggestion, you still need a lot of your old PR.

I was thinking something like using the existing enums, detecting if it was inverted or not from the name, and adding a flag to ValveGearConfig that's used similar to how you have it setup in previous PRs.

PR #1475, opened 2024-04-14
Fixed Grade Crossing Math

km00700:
Manipulated terms in Grade crossing math to reduce chained operations. Symmetrized behavior for positive and negative values of q to make grade crossing heights symmetrical on both sides of track.


cam72cam:
Hey @km00700, sorry it's taken me a while to get to this!  Have you tested with different gauges?  Where does 0.306 come from?

km00700:
I tested it on gauges from model to Ludicrous. (10m)
0.306 came from closely matching the original function in Desmos.

PR #1512, opened 2024-11-06
Remove decimals from GUI speed readout and add reverser keybinds to en_us.lang

SebastianD334:


SteampsonJake:
Oooh, that would be cool

PR #1521, opened 2025-02-17
[Feature]Make stock drop nothing when player is in creative mode

Goldenfield192:


src/main/java/cam72cam/immersiverailroading/entity/EntityBuildableRollingStock.java
```diff
@@ -362,6 +362,10 @@ public void onDamage(DamageType type, Entity source, float amount, boolean bypas
 		}
 
 		if (this.isDead() && shouldDropItems(type, amount)) {
+			if(source.isPlayer() && source.asPlayer().isCreative()){
```
cam72cam, line 365: We should probably make sure that this is a config option.  I could see someone in creating mode being confused by this default.

The problem is that this fires server-side and does not care about client settings.  We could add a server setting for this.  If it's a popular feature that folks want to turn on and off frequently, we could consider adding a gamerule, but that's not integrated into UMC yet.

cam72cam: Let's start off with it as a "common" setting as it would also be useful in singleplayer.

Goldenfield192: Thanks for the review and I've added it to Config/Balance as a setting

src/main/java/cam72cam/immersiverailroading/Config.java
```diff
@@ -222,6 +222,9 @@ public static List<ItemStack> getVillagerPayout() {
 
 		@Comment("Round to nearest bucket")
 		public static boolean RoundStockTankToNearestBucket = true;
+
+		@Comment("Does stock drops itself/components when player is in creative mode?")
```
cam72cam, line 226: I think we should put this in the debug section, it's more of a creative tweak.

Goldenfield192: Thanks and changed in commit [8a42be2](https://github.com/TeamOpenIndustry/ImmersiveRailroading/pull/1521/commits/8a42be23ab6defd89824b6ce5f98e7aca205c7d2)

src/main/java/cam72cam/immersiverailroading/entity/EntityBuildableRollingStock.java
```diff
@@ -362,6 +363,10 @@ public void onDamage(DamageType type, Entity source, float amount, boolean bypas
 		}
 
 		if (this.isDead() && shouldDropItems(type, amount)) {
+			if(getWorld().isServer && !Config.ConfigDebug.StockDropInCreativeMode && source.isPlayer() && source.asPlayer().isCreative()){
```
cam72cam, line 366: Remove the .isServer check here.  It should already only be running on the server.

Goldenfield192: > Remove the .isServer check here. It should already only be running on the server.

Fixed in commit [5378359](https://github.com/TeamOpenIndustry/ImmersiveRailroading/pull/1521/commits/5378359a43059b9207ddcd7a44a803e737d75b66)

src/main/java/cam72cam/immersiverailroading/Config.java
```diff
@@ -287,7 +286,10 @@ public static class ConfigDebug {
 		@Comment("Number of physics steps to cache for future movement / send in packets.  DO NOT CHANGE UNLESS YOU KNOW WHAT YOU ARE DOING")
 		@Range(min = 10, max = 60)
 		public static int physicsFutureTicks = 10;
-    }
+
+		@Comment("Does stock drops itself/components when player is in creative mode?")
+		public static boolean StockDropInCreativeMode = false;
```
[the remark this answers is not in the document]
Goldenfield192: Fixed in commit [52a8701](https://github.com/TeamOpenIndustry/ImmersiveRailroading/pull/1521/commits/52a87018ce9d86e9ea7a21ed7bedcc445e8dbcb6)

cam72cam:
Thanks @Goldenfield192 !

PR #1523, opened 2025-02-20
Chore: update GH actions script to v4

Goldenfield192:
With the deprecation of v1&v2 workflow at the beginning of 2024, IR's build pipeline had been exist in name only since it used v2.
This PR aims to migrate the build pipeline to modern versions of GH actions in order to recover its function.


cam72cam:
Thanks @Goldenfield192

PR #1524, opened 2025-02-25
Fix dimension change crash in 1.7.10

KellenHurrey:
Check to see if the player actually exists before checking what they are riding. Needs my pr for UMC to work.


src/main/java/cam72cam/immersiverailroading/ImmersiveRailroading.java
```diff
@@ -197,6 +197,9 @@ public void postRender(EntityMoveableRollingStock entity, RenderState state, flo
 				GlobalRender.registerItemMouseover(IRItems.ITEM_MANUAL, MBBlueprintRender::renderMouseover);
 
 				GlobalRender.registerOverlay((state, pt) -> {
+					if (MinecraftClient.getPlayer() == null){
```
cam72cam, line 200: The overlay should not fire if MinecraftClient.getPlayer() == null, let's move this check into UMC instead.

KellenHurrey: I'll add a pr in UMC to not continue in the event if `!MinecraftClient.isReady()`

PR #1527, opened 2025-03-08
Fix: Un-built stock can be controlled by dragging

Goldenfield192:
IR doesn't check stock's building status when processing control drag, so users can control an unfinished stock. This this pr fixes that.

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/entity/ClientPartDragging.java
+++ b/src/main/java/cam72cam/immersiverailroading/entity/ClientPartDragging.java
@@ -180,6 +180,10 @@ public class ClientPartDragging {
             Double min = null;
 
             for (EntityRollingStock stock : world.getEntities(EntityRollingStock.class)) {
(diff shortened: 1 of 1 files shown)


km00700:
I think stock that’s partially built should be able to be pulled. Might be good to check that it at least has wheels though.

Goldenfield192:
> I think stock that’s partially built should be able to be pulled. Might be good to check that it at least has wheels though.

No they can be pulled even the cab is not built which is weird

km00700:
Not really, most rolling stock doesn't have a cab. Unless you mean it can be "controlled" and not "pulled" in which case, it shouldn't have controls until it's fully built.

Goldenfield192:
> Not really, most rolling stock doesn't have a cab. Unless you mean it can be "controlled" and not "pulled" in which case, it shouldn't have controls until it's fully built.

I'll change the expression, thanks

PR #1536, opened 2025-04-10
fix: incorrect particle emitter position

Goldenfield192:
Diesel exhaust/Steam chimney's position will be incorrectly offseted based on speed, this pr fixes that


cam72cam:
The fun part is that this changes depending on what MC version you are on.  Can you try on both 1.12.2 and more recent MC versions?

Goldenfield192:
> The fun part is that this changes depending on what MC version you are on. Can you try on both 1.12.2 and more recent MC versions?

Tested in development environment of 1.14.4 and 1.16.5, both of them works

PR #1548, opened 2025-08-10
[Feature] Optimize CubicCurve

Goldenfield192:
Thanks to @cam72cam's advice, this PR partially overhauls the model algorithm to optimize performance and prevent track overlap issues caused by iterating from both sides of the curve.
Example:
Old method: 

New method:

Additionally, the generation code is now over 4x faster with the new method, depending on actual curve.


Goldenfield192:
> I loaded my existing "difficult track" world to test pathing and noticed that existing track does not render the last segment in a lot of scenarios:
> 

This should be fixed with newer commit

PR #1556, opened 2025-08-19
[feat] Added semi-transparent parts

poizzy:
This PR aims to add partial transparency to IR.  Every object with `ALPHA` in the name is drawn as a semi-transparent part. The implementation isn't ideal, it feels more like a temporary solution than anything near to permanent. There are also some bugs that I havn't been able to solve yet and don't know if they are even solvable (see: [here](https://discord.com/channels/355731184157720578/356567663788425216/1407052430922879077))


src/main/java/cam72cam/immersiverailroading/model/ModelState.java
```diff
@@ -300,9 +301,13 @@ public void render(OBJRender.Binding vbo, EntityMoveableRollingStock stock, List
 
         levels.forEach((level, litGroups) -> {
             List<String> animated = animatedGroups.isEmpty() ? Collections.emptyList() :
-                    litGroups.stream().filter(g -> animatedGroups.containsKey(g)).collect(Collectors.toList());
-            List<String> notAnimated = animatedGroups.isEmpty() ? litGroups :
-                    litGroups.stream().filter(g -> !animatedGroups.containsKey(g)).collect(Collectors.toList());
+                    litGroups.stream().filter(g -> animatedGroups.containsKey(g)).filter(g -> !g.contains("ALPHA")).collect(Collectors.toList());
```
cam72cam, line 304: All of these streams should be replaced with a single loop at this point for performance.

src/main/java/cam72cam/immersiverailroading/model/ModelState.java
```diff
@@ -225,25 +226,87 @@ private boolean hasGroupFlag(String group, String filter) {
...
+            if (this == obj) return true;
+            if (!(obj instanceof LightKey)) return false;
+            LightKey k = (LightKey) obj;
+            return sky == k.sky && block == k.block;
+        }
+    }
+
+    private static final BlendMode ALPHA_BLEND = new BlendMode(BlendMode.GL_SRC_ALPHA, BlendMode.GL_ONE_MINUS_SRC_ALPHA);
+    private final List<String> singletonList = new ArrayList<>();
+    private final Map<LightKey, RenderGroup> levels = new Object2ObjectOpenHashMap<>();
```
cam72cam, line 286: This does not need to be cached, the existing render pipeline creates a new ModelState frequently (for better or worse)

PR #1565, opened 2025-08-31
[Feature] Advanced track model definition

### LINKED ISSUE #1197: [suggestion] Multiple-model track types
#### Overview
Currently, in IR, a modeler who makes a track type does so by making a 1-block segment of track, which IR then repeats indefinitely. I suggest that in addition, a track type can be defined with a list of models, which are either repeated in a loop (for example, for OLE catenary), or picked at random (such as American narrow gauge, where each sleeper is at a random angle)

##### Example 1
```json
"repeat_style":"loop",
"track_models":{
  1:"track/meter_gauge_catenary/pole_left.obj",
  2:"track/meter_gauge_catenary/no_pole.obj",
  3:"track/meter_gauge_catenary/no_pole.obj",
  4:"track/meter_gauge_catenary/no_pole.obj",
  5:"track/meter_gauge_catenary/no_pole.obj",
  6:"track/meter_gauge_catenary/no_pole.obj",
  7:"track/meter_gauge_catenary/no_pole.obj",
  8:"track/meter_gauge_catenary/no_pole.obj"
}
```
In the above example, the block that the blueprint is placed on uses the `pole_left.obj` track piece, then the next seven segments use the `no_pole.obj` track piece, then it repeats again

Note the existence of the `"repeat_style"` parameter. This would default to `"loop"`, but could also be set to `"random"` (each track segment uses a random model from the list), or `"none"` (it goes through the items in sequence like `"loop"`, but when reaching the last model it just repeats that model forever, rather than going through the entire sequence again)

##### Example 2
```json
"repeat_style":"random",
"track_models":{
  1:"track/us_narrow_gauge/left_45.obj",
  2:"track/us_narrow_gauge/left_30.obj",
  3:"track/us_narrow_gauge/left_15.obj",
  4:"track/us_narrow_gauge/center.obj",
  5:"track/us_narrow_gauge/right_15.obj",
  6:"track/us_narrow_gauge/right_30.obj",
  7:"track/us_narrow_gauge/right_45.obj"
}
```
Each track segment would use a random track piece, in this case representing different angles of sleeper rotation. The "random" formula would use a seed derived from the x, y, z co-ords, the track type (straight, curve, custom curve, etc.), the direction the track was placed, and how far from the first track block this segment is.

##### Example 3
```json
"repeat_type":"none",
"track_models":{
  1:"track/level_crossing/fishplate.obj",
  2:"track/level_crossing/guide_rails_start.obj",
  3:"track/level_crossing/crossing_track_with_guide_rails.obj"
}
```

The first track segment (the one where it was placed) would use the `fishplate.obj` track piece, the second the `guide_rails_start.obj` track piece, and the third and all thereafter the `crossing_track_with_guide_rails.obj` track piece.

#### Associated stuff
In the track blueprint menu, there should be a toggle something along the lines of "use single model". This would use the first track piece for `"loop"` and `"random"` repeat types, and the final track piece for `"none"` repeat type.

Goldenfield192:
Closes #1197 
This PR adds:
1. `TABLE` Modifier for Track Model Parts: Allows track model parts to render on turntable/transfer table dynamic parts without being scaled, enabling the rendering of bridges.
2. Supports for track with multiply pieces
3. A fatal bug introduced in #1563 for not removing all anonymous inner classes

A manual could be found at [My gist](https://gist.github.com/Goldenfield192/70bc96cce31cd1a784868fbe302073b5)


src/main/java/cam72cam/immersiverailroading/track/BuilderIterator.java
```diff
@@ -222,14 +223,15 @@ public List<VecYawPitch> getRenderData() {
...
+					vec.addChild(new VecYawPitch(switchPos.x, switchPos.y, switchPos.z, switchPos.yaw, switchPos.pitch, (1 - angle / 180) * (float) scale, TrackModelPart.RAIL_LEFT));
+					vec.addChild(new VecYawPitch(cur.x, cur.y, cur.z, cur.yaw, cur.pitch, (1 + angle / 180) * (float) scale, TrackModelPart.RAIL_RIGHT));
 				} else {
-					data.add(new VecYawPitch(cur.x, cur.y, cur.z, cur.yaw, cur.pitch, (1 - angle / 180) * (float) scale, "RAIL_LEFT"));
-					data.add(new VecYawPitch(switchPos.x, switchPos.y, switchPos.z, switchPos.yaw, switchPos.pitch, (1 + angle / 180) * (float) scale, "RAIL_RIGHT"));
+					vec.addChild(new VecYawPitch(cur.x, cur.y, cur.z, cur.yaw, cur.pitch, (1 - angle / 180) * (float) scale, TrackModelPart.RAIL_LEFT));
+					vec.addChild(new VecYawPitch(switchPos.x, switchPos.y, switchPos.z, switchPos.yaw, switchPos.pitch, (1 + angle / 180) * (float) scale, TrackModelPart.RAIL_RIGHT));
 				}
-				data.add(new VecYawPitch(cur.x, cur.y, cur.z, cur.yaw, cur.pitch, (float) scale, "RAIL_BASE"));
+				data.add(vec);
```
cam72cam, line 234: You do this so the "component" here is styled consistently?  I think that makes sense.

Goldenfield192: Here is to make the size of `renderData` matches pieces count to parse `order`

src/main/java/cam72cam/immersiverailroading/model/TrackModel.java
```diff
@@ -1,45 +1,298 @@
...
-                groups.add(group);
+        this.groupNamesMapper = new HashMap<>();
+        if(isSingle) {
+            Map<TrackModelPart, List<String>> groups = new HashMap<>();
+            for(TrackModelPart part : TrackModelPart.values()){
+                List<String> parts = this.groups().stream().filter(part::is).collect(Collectors.toList());
+                groups.put(part, parts);
             }
+            groupNamesMapper.put("single", groups);
+            randomMap.put("single", () -> "single");
```
cam72cam, line 43: Is this necessary?  Shouldn't the order look up the group directly?

cam72cam: Ah, you treat random as the main lookup and add singleton entries below as well.  Clever :+1:

src/main/java/cam72cam/immersiverailroading/model/TrackModel.java
```diff
@@ -1,45 +1,298 @@
...
+        matrix.scale(scale, scale, scale);
+
+        List<String> tables = groupNames.get(TrackModelPart.TABLE);
+
+        if(piece.getParts().contains(TrackModelPart.TABLE)){
+            builder.draw(tables, matrix);
+        }
+
+        if (piece.getLength() != -1) {
+            matrix = matrix.copy().scale(piece.getLength() / info.settings.gauge.scale(), 1, 1);
```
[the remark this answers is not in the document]
Goldenfield192: Solved with [a13d98e](https://github.com/TeamOpenIndustry/ImmersiveRailroading/pull/1565/commits/a13d98e7b0627cb200db338dc64e3b5b48dae61d)

PR #1567, opened 2025-09-03
snowLayers property for rolling stock

FnordyCave:
Allows modelers to set how many snow layers should be left uncleared. 
First time working on a public codebase so feel free to criticize


cam72cam:
This is great, thanks!

Issue #1382, opened 2023-06-29
Fuel categories (for internal combustion engines)


CherubAgent1440:
**Describe the suggestion**  
Basically, a pack should be able to define a "Fuel category" - a list of liquids that can be used as "diesel fuels" for specific vehicles, combined with their efficiencies. A specific fluid can be in multiple fuel categories with different efficiencies in each (e.g. kerosene can be used in a gasoline engine, but it is a more efficient use of the kerosene to put it in a kerosene engine)

A locomotive can then select a fuel category in the json.

Built-in fuel categories should include "diesel" (diesel [high efficiency], biodiesel [medium efficiency]) and "gasoline" (gasoline [high efficiency], ethanol [low efficiency], kerosene [medium efficiency]). Maybe also "kerosene" (kerosene [high efficiency], ethanol [medium efficiency]).  
> *Note - if you know more about the relative efficiency of kerosene, gasoline, and ethanol in kerosene and gasoline engines, feel free to use your values over mine. I'm not very knowledgeable in these things.*

The mod config should also define "universal fuels", which would include things that are just named "fuel" or "fuelium" or whatever. Maybe this should include lava at a low efficiency (for vanilla compatibility)

If two packs define a fuel type with the same ID then the list of fluids are merged, keeping just the higher efficiency instance of any duplicate entries (So if I defined a `magic` fuel category that used `liquid_xp` at efficiency 200 and `lava` at efficiency 100, whilst another pack I have installed made a `magic` fuel category that used `lava` at efficiency 150 and `liquid_mana` at efficiency 250, then the full list is `liquid_xp` efficiency 200, `lava` efficiency 150, and `liquid_mana` efficiency 250).

Additionally, there should be a config setting that allows you to state that two fuel category IDs are the same (treated as if they had the same name) (continuing from the above example, a third pack might define an `arcane` fuel category, and since this is equivilant to the `magic` above, then the user may wish to tweak the config to make them equivalent to each other)

Finally, there should be a config option whereby the user can add liquids to specific fuel categories (e.g. if I had a mod that added a liquid that is internally called "octane" I would want it to run in a gasoline or kerosene engin, but might not want to make a whole pack to add it).

Oh and there should be a config option of "all engines use all fuels" which basically makes all fuel types that exist in the game be combined as if they had the same name.

[...]

PR #1577, opened 2025-09-12
[Feature] Add ability to customize diesel fuel by definition

### LINKED ISSUE #1466: Diesel Locomotive Fuel Whitelist
**Describe the suggestion**  
JSON tag for diesel locomotives so that diesels can only use fuels specified in the JSON, as opposed to the main list in the config. (If the tag isn't present the loco would just use the main list)
For example, it the JSON it might state "allowed_fuels": gasoline, and the locomotive would only be allowed to use gasoline.

**Applications of the suggestion**  
This allows locomotives to use only certain fuels, without said fuel being available to all locos. For instance, an experimental hydrogen locomotive without allowing all diesels in game to be able to use hydrogen as a fuel.

Goldenfield192:
Resolves #1466, #1382 

This PR gives modeler the ability to lock down which fuels a diesel locomotive accepts and set custom burn times for them right in the stock properties. Add a `fuel_override` list inside `properties`, and the locomotive will ignore anything that isn't on that list.

Each entry follows the pattern `fluid_name` or `fluid_name:burn_time`.

- If a fluid is already defined in the main config, you can simply name it to use its default burn time, or append `:burn_time` to override that value for this locomotive.

- For a fluid not defined in the main config, you must provide a burn time directly; otherwise it will be skipped.

For instance,

```json
...
"properties": {
    ...
    "fuel_override": [
        "water:1000",
        "ethanol:1700",
        "gasoline",
        "lava"
    ]
    ...
},
...
```
In this example configuration:
- The locomotive will only accept `water`, `ethanol`, and `gasoline` as valid fuels.
- `water` uses a custom burn time of `1000` (not defined in the main config, specified directly here).
- `ethanol` overrides the default burn time for this locomotive from the main config with a new value of `1700`.
- `gasoline` falls back to using its predefined burn time (`100` as defined in the main config).
- `lava` will be skipped in default environment because it neither exist in default config nor have burntime specified here

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/entity/LocomotiveDiesel.java
+++ b/src/main/java/cam72cam/immersiverailroading/entity/LocomotiveDiesel.java
@@ -16,8 +16,7 @@ import cam72cam.mod.fluid.Fluid;
 import cam72cam.mod.fluid.FluidStack;
 import cam72cam.mod.serialization.TagField;
 
-import java.util.List;
-import java.util.OptionalDouble;
+import java.util.*;
 
 public class LocomotiveDiesel extends Locomotive {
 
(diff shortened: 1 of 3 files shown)


Goldenfield192:
> Is there a reason you don't want to use a more traditional structure here:
> 
> ```json
> ...
> "properties": {
>     ...
>     "fuel_override": [
>         "water":  1000,
>         "ethanol": 1700,
>         "gasoline": 0,
>         "lava": 0
>     ]
>     ...
> },
> ...
> ```

If we want to use a defined fluid and let it be 0 may seem weird here

Issue #1493, opened 2024-06-21
Add a way to filter which set of doors open using the door augment


ZeBoss07:
**Describe the suggestion**  
Have a text box or something pop up when right clicking the door actuator to set the name of the CG of the doors wanting to be opened, allowing only one side of doors to open

**Applications of the suggestion**  
Automatic door opening onto a platform and not have the other side open onto a wall/running track

Goldenfield192:
Resolved in #1578

Issue #1038, opened 2021-08-11
Consider locomotive / stock in-game tagging system (in addition to existing OC/CC)


cam72cam:
habatman in discord suggested that a tagging system that could be read by switches / augments could be useful.  Now that most augments respond to redstone, I think interacting with the tagging system via some sort of tool and detector augment could be neat.  

This will require some new GUI work and a fair amount of programming.

PR #1578, opened 2025-09-12
[Feature] Add stock tag system & augment GUI

### LINKED ISSUE #1459: An overhaul for the augments filter system
The idea i have is so when placing down an augment, using a new item (Clipboard or some sort of writing thing) it stores filters that can be applied to the augments, this can vary from a specific piece of roiling stock, to multiple stocks, types of stock (like only locos or only coaches) specific items or entity that reside in the cars, and even a blacklist mode that dose the opposite.

This feature would revolutionize automation for IR with the ability to set paths for certain trains, as-well as fix loaders taking the locos fuel and cars being loaded that you don't want loaded.

Goldenfield192:
Resolves #1459, #1493, #1038 
### Augment GUI and Filtering Overhaul

The rail augment now has its own dedicated GUI, bringing all the old augment settings together.  
Door actuators also get a new **“Door Actuator CG filter”** field — use it to limit control to doors that share a specific CG value.

We've rebuilt the stock filtering system from scratch. These are the filter patterns you can now write:

- `type:[stock type]` — matches by stock type
- `tag:[stock tag]` — matches by tag
- `stock:[name in stock.json]` — matches a specific stock definition (old `augmentFilterID` entries migrate here automatically)
- `works:[locomotive works]` — matches by builder or works
- `author:[modeler name]` — matches by author
- `pack:[pack name]` — matches by pack
- `nametag:[nametagged stock name]` — matches nametagged stock (old `stockTag` entries migrate here automatically)

You can combine filters with `&&` (and), `||` (or), and parentheses, just like in most C‑like languages.  
Fill in the **Included Tags** field to pick which stocks you want; leave it empty and all stocks match. Then use **Excluded Tags** to drop specific stocks from the list — leave it empty and nothing gets removed.

**Example:** To work on everything from *General Electric* or *Union Pacific Railroad* except the *Class B‑B‑70/70‑4HM829 Boxcab*, set it up like this:

> Included Tags: `works:General Electric || works:Union Pacific Railroad`  
> Excluded Tags: `stock:ge_40_ton_boxcab`

The old behaviour of clicking a name tag or a stock to set a filter is gone. All filtering now happens right inside the GUI, using the text patterns shown above.


src/main/java/cam72cam/immersiverailroading/items/ItemRollingStock.java
```diff
@@ -138,31 +138,15 @@ public List<String> getTooltip(ItemStack stack)
...
-					}
-					return ClickResult.ACCEPTED;
-				default:
-					break;
-				}
-			}
-		}
+//		if(world.isServer && BlockUtil.isIRRail(world, pos)) {
+//			TileRailBase base = world.getBlockEntity(pos, TileRailBase.class);
+//			//TODO
```
cam72cam, line 143: Looks like this part still needs to be implemented

Goldenfield192: Wait I almost forgot

src/main/java/cam72cam/immersiverailroading/tile/TileRailBase.java
```diff
@@ -46,8 +48,14 @@ public class TileRailBase extends BlockEntityTrackTickable implements IRedstoneP
...
 	@TagField("augment")
 	private Augment augment;
+	@Deprecated
 	@TagField("augmentFilterID")
 	private String augmentFilterID;
+	@TagField("positive_filter")
+	private String positive;
+	@TagField("negative_filter")
+	private String negative;
+	private List<Predicate<EntityRollingStock>> compiledFilter = new LinkedList<>();
```
cam72cam, line 58: I think the filter compiler should result in a single predicate, as they are easy to chain.

src/main/java/cam72cam/immersiverailroading/items/ItemRollingStock.java
Goldenfield192: Removed

src/main/java/cam72cam/immersiverailroading/tile/TileRailBase.java
Goldenfield192: Fixed since [e452e13](https://github.com/TeamOpenIndustry/ImmersiveRailroading/pull/1578/commits/e452e13e1284395d697855f7327e6769e7c859ab), stable version at [b096e19](https://github.com/TeamOpenIndustry/ImmersiveRailroading/pull/1578/commits/b096e192c37e4006994800aad48ace8aa0e14316)

PR #1579, opened 2025-09-12
feat: add `whitelistBlocks` to prevent destruction to certain blocks

### LINKED ISSUE #1427: Break Block whitelist
a whitelist for the break blocks setting, enabling it to break certain kinds of block (e.g. snow or leaves) while not breaking other blocks.
this would enable particularly narrow gauge trains to clear their paths without also damaging platforms in a station.

Goldenfield192:
Resolves #1427 

This PR adds a config entry to mark certain types of block undestroyable by train

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/Config.java
+++ b/src/main/java/cam72cam/immersiverailroading/Config.java
@@ -43,11 +43,6 @@ public class Config {
 			// Other
 			ConfigBalance.dieselFuels.put("olive_oil", 40);
 		}
-
-		if (ConfigDamage.whitelistBlocks.isEmpty()) {
-			//Little tiles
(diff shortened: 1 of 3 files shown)


src/main/java/cam72cam/immersiverailroading/Config.java
```diff
@@ -60,6 +60,11 @@ public static class ConfigDamage {
 		@Comment("Trains should break block")
 		public static boolean TrainsBreakBlocks = true;
 
+		@Comment("Which block is reserved if TrainsBreakBlocks is true")
+		public static String[] whitelistBlocks = new String[]{
```
cam72cam, line 64: Let's use a more specific name here like "TrainsIgnoreBlocks"

PR #1582, opened 2025-09-14
[Fix] Revert #1555

### LINKED ISSUE #1581: Multiblock creating issue
After #1555 merged at 1.16.5 creating multiblock will cause server freeze

Goldenfield192:
Resolves #1581 

### Background
In #1555 multiblock creation was moved to multi-thread model. However, in Minecraft 1.16.5, all chunk read operations are forced to execute on the main thread, which breaks the foundation of this feature and resulting in the game to freeze(then crash) with #1555 .

This PR reverts it to prevent such crash.


Goldenfield192:
Otherwise we'll have to use mixin

PR #1592, opened 2025-09-29
Adjusted hp, added ps, adjusted default config, organized unit library

SebastianD334:
Added new unit: PS (metric horsepower)
Adjusted hp to W conversion factor from metric (735.5 W) to imperial (745.7 W).
Updated and reorganized locomotive.caml and steam.caml files to improve legibility and order.
Adjusted tooltips with correctly capitalized kg and Tractive Eff. instead of Traction, as well as adding spaces between numerators and units for power, tractive effort, and speed.
Reordered and adjusted defaults for units in config so metric is the default for all of them (km/h, kW, kN, bar, °C) and they're all ordered metric, SI, imperial (except power, where PS is last, which puts it between imperial horsepower and metric in a way).
Adjusted some comments in the config to be more clear or accurate.

(The commit name includes bar because I misremembered that it was actually added in the previous PR)

I apologize for not splitting this into multiple commits, but everything was really intertwined.

### DIFF
--- a/src/main/resources/assets/immersiverailroading/lang/en_us.lang
+++ b/src/main/resources/assets/immersiverailroading/lang/en_us.lang
@@ -116,16 +116,15 @@ gui.immersiverailroading:track.place_blueprint_true=Place:      Blueprint
 gui.immersiverailroading:track.place_blueprint_false=Place:      Track
 
 gui.immersiverailroading:loco.works=Works:          %s
-gui.immersiverailroading:loco.horse_power=Horse Power: %s
 gui.immersiverailroading:loco.power=Power:          %s
-gui.immersiverailroading:loco.tractive_effort=Traction       %s
+gui.immersiverailroading:loco.tractive_effort=Tractive Eff:  %s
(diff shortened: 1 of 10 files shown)


src/main/resources/assets/immersiverailroading/lang/en_us.lang
```diff
@@ -116,16 +116,15 @@ gui.immersiverailroading:track.place_blueprint_true=Place:      Blueprint
 gui.immersiverailroading:track.place_blueprint_false=Place:      Track
 
 gui.immersiverailroading:loco.works=Works:          %s
-gui.immersiverailroading:loco.horse_power=Horse Power: %s
 gui.immersiverailroading:loco.power=Power:          %s
-gui.immersiverailroading:loco.tractive_effort=Traction       %s
+gui.immersiverailroading:loco.tractive_effort=Tractive Eff:  %s
```
Goldenfield192, line 120: I'd recommend using full name instead of abbreviation

PR #1593, opened 2025-10-04
[Refactor] Cleanup and add pagination for TrackGUI

Goldenfield192:
This PR cleans current TrackGui and add ability to add more pages to define more properties


cam72cam:
Would you mind adding a screenshot of the two pages at different scales?

PR #1598, opened 2025-10-06
[feat] Multi-floor capability

poizzy:
2nd time the charme because I fucked up...


src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java
```diff
@@ -77,8 +78,8 @@ public abstract class EntityRollingStockDefinition {
     private double rearBounds;
     private double heightBounds;
     private double widthBounds;
-    private double passengerCompartmentLength;
-    private double passengerCompartmentWidth;
+    @Nullable public Double passengerCompartmentLength;
+    @Nullable public Double passengerCompartmentWidth;
```
Goldenfield192: `@Nullable` annotation may cause building failure on 1.7.10

src/main/java/cam72cam/immersiverailroading/util/VecUtil.java
```diff
@@ -61,4 +62,33 @@ public static Vec3d fromWrongYawPitch(float distance, float rotationYaw, float r
 	public static Vec3d between(Vec3d front, Vec3d rear) {
 		return new Vec3d((front.x + rear.x) / 2, (front.y + rear.y) / 2, (front.z + rear.z) / 2);
 	}
+
+	public static Vec3d removePitch(Vec3d vec, double pitchDegrees) {
```
Goldenfield192: Maybe we should merge this and `reapplyPitch` into `rotatePitch`?

src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java
```diff
@@ -675,33 +656,31 @@ private static class HeightMapData {
                     .collect(Collectors.toList());
             data = new float[components.size() * xRes * zRes];
 
-            VertexBuffer vb = def.model.vbo.buffer.get();
+            FaceAccessor visitor = def.model.getFaceAccessor();
```
cam72cam: Using the FaceAccessor / the changes here are about 8% slower in my testing.  The code is much cleaner, so I'd be willing to live with it.

cam72cam (changes requested):
Overall, the multi-floor stuff feels really smooth!

I need the following fixed before merge:
* Walking between cars does not work, see comment
* All movement stops when colliding with a wall, this should try to smear instead
* Gauge scaling of models is completely broken

Jeronimo97:
I have some feedback from testing this PR. It fucks up the RAM.

PR #1613, opened 2025-11-01
[Fix] Rail bed z-fighting(#1612)

### LINKED ISSUE #1612: Transfer Table/Turntable bed model z-fighting
Image

As the picture above, the upper layer of model will have a z-fighting with lower models

Goldenfield192:
Resolves #1612 

This PR changes Transfer Table/Tuentable rail bed render logic to avoid z-fighting on edges of model (Only available to new placed tracks)


Goldenfield192:
> Could we add a migration for this between TE versions?

I think we couldn't as...
1) We can't determine the loading order of a block and its parent, thus not able to migrate all at once;
2) We can hardly distinguish if a tile is actually a bottom part of table, introducing migration may cause unwanted behaviours

PR #1621, opened 2025-11-20
[Draft Feature] Track snapping

Goldenfield192:
This PR introduces the feature that track blueprint/golden spike can snap to close track ends, with angle applied

To be discussed:
- Should we make it toggled by shift-clicking the air instead of config?
- If possible, could we add `TrackNode` to replace `PlacementInfo` then build off-world storage?


Goldenfield192:
Given of the merge of #1643 this PR requires a huge overhaul to work with main branch

PR #1624, opened 2025-11-29
[Feature] Enhanced cargo component

### LINKED ISSUE #1491: Alternative method of loading and unloading items
**Describe the suggestion**  
I would like to suggest different method for loading and unloading items on wagons.

Loading items would be done by throwing the items from above onto the hitbox of the wagon.
- this option could be enabled/disabled in the wagon's config **"hopper_on_top": "true/false",**

https://github.com/TeamOpenIndustry/ImmersiveRailroading/assets/68654943/de6abb2d-42de-44de-9b70-0089eaaebd3d

Unloading items would be done by widgets and "UNLOADING_POINT"
- "UNLOADING_POINT" would define where the items are dropped (small mesh on the model from which items are supposed to fall out)
- example WIDGET_1_CG_unload and UNLOADING_POINT_CG_unload

https://github.com/TeamOpenIndustry/ImmersiveRailroading/assets/68654943/1a9466c1-abcb-40ee-8eec-43c8ab6079cc

_Footage above shows only the concept._

**Applications of the suggestion**  
It would add some imersion and I think it would be a good alternative to the augments that are to be removed/change in the future.

Goldenfield192:
Resolves #1491 

This PR makes `CARGO_FILL` and `CARGO_ITEMS` able to load dropped entities, scanning every 5 ticks, and adds a config setting to disable it

As profiled with 300 stocks, old method cost 800ms in 30s period and new method costed 1100ms, which is reasonable to me

### DIFF
--- a/src/main/java/cam72cam/immersiverailroading/Config.java
+++ b/src/main/java/cam72cam/immersiverailroading/Config.java
@@ -97,6 +97,9 @@ public class Config {
 
 		@Comment("Enable coupler slack")
 		public static boolean slackEnabled = true;
+
+        @Comment("Enable CARGO_FILL and CARGO_ITEMS load items dropped to the world")
+        public static boolean allowCargoLoadDroppedItem = true;
     }
 
 	@Name("balance")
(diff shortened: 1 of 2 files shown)


cam72cam:
Given that this is easy to turn off, it seems reasonable to me

PR #1666, opened 2026-07-10
[feat] Add multi-floor compatibility

poizzy:
I think everyone already knows what this PR adds, so I'll skip the description. However I added one slight improvement in comparison to the old PR: Sliding along walls.


alexkubel:
Yippee!

src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java
```diff
@@ -580,34 +596,154 @@ public List<ModelComponent> getComponents(ModelComponentType name) {
...
-        pos = new Vec3d(pos.x, passengerCenter.y - (shouldSit ? 0.75 : 0), pos.z + passengerCenter.z);
+        return null;
+    }
+
+    public Vec3d correctPassengerBounds(Gauge gauge, Vec3d passengerOffset, boolean shouldSit) {
+        // Flip coords
+        passengerOffset = passengerOffset.rotateYaw(-90);
+
+        // TODO: check for loop
+        for (float searchRange : new float[]{0.5f, 5f, 10f}) {
```
Goldenfield192: If the player is newly mounted then there may be more than 10m between him and nearest `FLOOR`, could we add a flag here to check that condition (and use the whole mesh to search)?

src/main/java/cam72cam/immersiverailroading/floor/NavMesh.java
```diff
@@ -0,0 +1,286 @@
...
+        return String.format(Locale.ROOT, "%.4f,%.4f,%.4f", v.x, v.y, v.z);
+    }
+
+    public Edge closestBoundaryEdge(Vec3d point) {
+        Edge closest = null;
+        double closestDistSq = Double.MAX_VALUE;
+        for (Edge edge : floorBoundaryEdges) {
+            Vec3d onSeg = closestPointOnSegment(point, edge.start, edge.end);
+            double dx = point.x - onSeg.x, dy = point.y - onSeg.y, dz = point.z - onSeg.z;
+            double distSq = dx*dx + dy*dy + dz*dz;
```
Goldenfield192: Could we turn to `Vec3d.distanceToSquared` here as it has an identical effect, to ease others' cognitive load?

src/main/java/cam72cam/immersiverailroading/floor/NavMesh.java
```diff
@@ -0,0 +1,286 @@
...
+            if (distSq < closestDistSq) {
+                closestDistSq = distSq;
+                closest = edge;
+            }
+        }
+        return closest;
+    }
+
+    private static Vec3d closestPointOnSegment(Vec3d p, Vec3d a, Vec3d b) {
```
Goldenfield192: If the method is only called on startup, could we use some util methods to add semantics? Like, `Math.clamp` and flow methods from `Vec3d`

Goldenfield192: Uh just ignore this

src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java
```diff
@@ -580,34 +596,154 @@ public List<ModelComponent> getComponents(ModelComponentType name) {
...
 
-        if (Math.abs(pos.x) > this.passengerCompartmentWidth / 2 * gs) {
-            pos = new Vec3d(Math.copySign(this.passengerCompartmentWidth / 2 * gs, pos.x), pos.y, pos.z);
+        if (intersects) {
+            Vec3d p1 = intersectingDoor.part.min;
+            Vec3d p2 = intersectingDoor.part.max;
+
+            p2 = new Vec3d(p2.x, p1.y, p2.z);
+
+            return p2.subtract(p1).normalize();
```
Goldenfield192, line 649: Should we treat doors as "planes" when calculating? This method seems not quite happy with thick doors, as well as on non-standard gauges

src/main/java/cam72cam/immersiverailroading/library/ModelComponentType.java
```diff
@@ -119,6 +119,10 @@ public enum ModelComponentType {
 	BRAKE_PRESSURE_X("BRAKE_PRESSURE_#ID#"),
 	COUPLED_X("COUPLED_#ID#"),
 
+	// Floor
+	FLOOR("FLOOR"),
```
Goldenfield192, line 123: Could we add them in `shouldRender`, to prevent them from rendering when holding & rendering icons?

src/main/java/cam72cam/immersiverailroading/registry/EntityRollingStockDefinition.java
```diff
@@ -580,34 +596,154 @@ public List<ModelComponent> getComponents(ModelComponentType name) {
...
+//            if (t != null) {
+//               return true;
+//            }
+//        }
+//
+//        return false;
+//    }
+
+    // TODO Rename
+    public Vec3d correctMovement(EntityRollingStock stock, Gauge gauge, Vec3d passengerOffset, Vec3d movement) {
```
Goldenfield192: This seems broken on upstair/slope case, could we consider a fix?

Goldenfield192 (approved):

PR #1682, opened 2026-08-04
[feature or fix?] make track exchanger able to swap block by using secondary hand

Deepseasaltyfish:


src/main/java/cam72cam/immersiverailroading/items/ItemTrackExchanger.java
```diff
@@ -47,6 +50,22 @@ public void onClickAir(Player player, World world, Player.Hand hand) {
...
+		if (world.isServer && hand == Player.Hand.SECONDARY) {
+			ItemStack blockinfo = world.getItemStack(pos);
+			ItemTrackExchanger.Data data = new ItemTrackExchanger.Data(stack);
+			data.railBed = blockinfo;
+			data.write();
+			player.setHeldItem(Player.Hand.SECONDARY, stack);
+			return ClickResult.ACCEPTED;
+		}
+
+		return ClickResult.ACCEPTED;
```
Goldenfield192: Should we return `PASS` here as it is unhandled?

Deepseasaltyfish: done

